PR #3, opened 2021-08-20
Add mongo db

sarahmcdougall:
# Summary
This PR sets up the code to connect our server to an empty Mongo database and adds a script that deletes the contents of all the collections in the `deqm-test-server` database.

## New behavior
Our `deqm-test-server` now connects to an empty Mongo database, which allows us to create collections and input/delete data contained in these collections. When the server is run, the MongoDB connection is established with the configured `deqm-test-server` database. The actual database queries will be integrated in future tasks. A script was added to delete all the contents from all the collections in the database. 

Note: We do not need a script to initialize an empty MongoDB database, as MongoDB does this automatically.

## Code changes
The `src/util/mongo.js` file establishes the local host url, client, and database name (`deqm-test-server`). The `src/index.js` file imports this information as `mongoUtil` and connects to the Mongo database. `src/index.js` also contains a `config` variable that configures a simple patient profile. This was added so that we do not receive an `Error: No profiles configured.` The `config` variable tells the server where to locate our patient service and that we intend to support it on R4 - it corresponds to the `patient.service.js` file that was created in this PR. 

The `deleteCollections.js` file was created to allow us to delete all the contents from every collection in our database. It stores all our collections in a variable and then maps an async function to delete the contents from each collection. If desired, we can create a future task to delete the collections themselves.

# Testing guidance
First, visit https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/ to install MongoDB Community Edition on macOS, if it has not been done already. Then, run `npm install`.

To Run MongoDB as a macOS service, use the following command: `brew services start mongodb-community@5.0`. To stop running, use `brew services stop MongoDB-community@5.0`. You can verify that MongoDB is running by running `brew services list`. 

After you get MongoDB running, run `node index.js` from `src` and check that you receive a message that you have connected successfully to the FHIR server and the database. Try running `mongo` to check that you can access the MongoDB shell. 

Next, check that you can add content into the database using the MongoDB shell. For example, you can use the following commands:
`use deqm-test-server`
`db.test.insertOne({‘test-insertion’:true})`

This will insert the object into the `test` collection. You can use `db.test.find({})` to verify that the object was inserted. 

To delete the contents, run `node src/scripts/deleteCollections.js` in the terminal. Then run `mongo` to return to the MongoDB shell, and run `use deqm-test-server`, followed by `db.test.find({})`. You should not receive any output, as the contents have been deleted inside the `test` collection.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,6 +56,11 @@
           "resolved": "https://registry.npmjs.org/astral-regex/-/astral-regex-1.0.0.tgz",
           "integrity": "sha512-+Ryf6g3BKoRc7jfp7ad8tM4TtMiaWvbF/1/sQcZPkkS7ag3D5nMBCe2UfOTONtAkaG0tO0ij3C5Lwmf1EiyjHg=="
         },
+        "bson": {
+          "version": "1.1.6",
+          "resolved": "https://registry.npmjs.org/bson/-/bson-1.1.6.tgz",
+          "integrity": "sha512-EvVNVeGo4tHxwi8L6bPj3y3itEvStdwvvlojVxxbyYfoaxJ6keLgrTuKdyfEAszFK+H3olzBuafE0yoh0D1gdg=="
+        },
         "debug": {
           "version": "4.3.2",
           "resolved": "https://registry.npmjs.org/debug/-/debug-4.3.2.tgz",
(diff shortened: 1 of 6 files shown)


natjoe4 (approved):
This looks great. All (very well) described functionality works and the code looks quite clean. Don't imagine we'd need another reviewer on this, but I'll leave that up to you.

hossenlopp (approved):
Looks good. I think in the future we may want a script for building the collections and setting up indexes. Could also make a `db:reset` npm run script to make it reset the db and create everything easily. But we can make that as we establish all the stuff actually that interacts with the database.

PR #4, opened 2021-08-23
Crud.

natjoe4:
# Summary
We can now create, delete, and update documents in our mongo db as well as search our db for a document by id. 
## New behavior
After starting mongo db and running the server, we can now use HTTP requests to create, update, search for, and remove documents from the 'Patients' collection of our database
## Code changes
Added the following functions to util/mongo.controller.js
- createResource: adds a new document with the passed in data to the specified collection 
- updateResource: uses passed in data to update the document with the specified id
- findResourceById: returns the document with the specified id
- removeResource: deletes the document with the specified id from the collection
Added the following functions to services/base.service.js
- baseCreate: wraps the createResource function and passes in the appropriate data
- baseUpdate: wraps the updateResource function and passes in the appropriate data
- baseSearchById: wraps the findResourceById function and passes in the appropriate data
- baseRemove: wraps the removeResource function and passes in the appropriate data
Added the following functions to services/patient.service.js
- create: wraps the baseCreate function and passes in the appropriate data
- - can be called using a POST request with the url _{BASE_URL}_/patient
- update: wraps the baseUpdate function and passes in the appropriate data
- - can be called using a PUT request with the url _{BASE_URL}_/patient/_{id}_
- searchById: wraps the baseSearchById function and passes in the appropriate data
- - can be called using a GET request with the url _{BASE_URL}_/patient/_{id}_
- remove: wraps the baseRemove function and passes in the appropriate data
- - can be called using a DELETE request with the url _{BASE_URL}_/patient/_{id}_
# Testing guidance
A couple disclaimers to start things off: 
- The only HTTP request which will return data to the user is the GET request. Don't expect the other types of requests to give output unless there is an error
- The ids should all be strings when passed in. The search, update, and delete functions will only work for string ids, but the create request will allow you to insert ids of any type for now.
- don't confuse "id" (the custom field you create) with "_id" (the primary key mongo db autogenerates). the update, search, and delete requests all use id to locate documents
- id is NOT a primary key and therefore the database will allow you to create multiple documents with the same id for now. Probably best not to do this in testing as it can get confusing quickly.

_*Whew*_

"First, visit https://docs.mongodb.com/manual/tutorial/install-mongodb-on-os-x/ to install MongoDB Community Edition on macOS, if it has not been done already. Then, run npm install.

To Run MongoDB as a macOS service, use the following command: brew services start mongodb-community@5.0. To stop running, use brew services stop MongoDB-community@5.0. You can verify that MongoDB is running by running brew services list." - _McDougall, Sarah. "Add Mongo DB". [https://github.com/projecttacoma/deqm-test-server/pull/3](https://github.com/projecttacoma/deqm-test-server/pull/3) GitHub August 20, 2021_ 

For testing, I recommend downloading [Insomnia](https://insomnia.rest/pricing) to make requests and [robomongo](https://robomongo.org/) to view the mongo database without using terminal commands. Then,

1. Run `node index.js` from the src directory
2. Open robomongo and add a new connection. Mine was at localhost 27017 (not sure if this is typical)
3. Open Insomnia, set the request method to POST and enter the url http://localhost:3000/4_0_0/patient
4. Add a json object to the request body. This can really be anything so long as it has an id field which is set as a **STRING**. If you're feeling uninspired, here's what I used: `{
	"id": "1",
	"name": "test"
}`
5. Submit the request, then refresh your robomongo window (right click on New Connection -> refresh), navigate to deqm-test-server -> Collections -> Patient and ensure your new patient object appears
6. Change the request type to GET, and the url to http://localhost:3000/4_0_0/patient/_{id}_ . The response body should include all the information on the document you just submitted, including an _id field which is autogenerated by mongo
7. Change the request type to PUT, and update your request body with a new name. I used "mokeefe" out of respect and you should to.
8. Submit the request, then refresh your robomongo window to ensure the update went through on your document
9. Change the request type to DELETE. If you have liquid around you, pour it out, then submit, refresh your robomongo page and ensure "mokeefe" is deleted from your database.
10. Try adding multiple documents and make sure the updates and deletes edit the correct entries, and the searchById always retrieves the intended document.

### DIFF
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -1,15 +1,15 @@
 const { findResourceById, createResource, removeResource, updateResource } = require('../util/mongo.controller');
 
-const baseCreate = async (data, resourceType) => {
-  return await createResource(data, resourceType);
+const baseCreate = async ({ req }, resourceType) => {
+  return await createResource(req.body, resourceType);
 };
-const baseSearchById = async (id, resourceType) => {
-  return await findResourceById(id, resourceType);
+const baseSearchById = async (args, resourceType) => {
+  return await findResourceById(args.id, resourceType);
 };
-const baseUpdate = async (id, data, resourceType) => {
-  return await updateResource(id, data, resourceType);
+const baseUpdate = async (args, { req }, resourceType) => {
+  return await updateResource(args.id, req.body, resourceType);
 };
-const baseRemove = async (id, resourceType) => {
-  return await removeResource(id, resourceType);
+const baseRemove = async (args, resourceType) => {
+  return await removeResource(args.id, resourceType);
 };
 module.exports = { baseCreate, baseSearchById, baseUpdate, baseRemove };
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -1,0 +1,15 @@
+const { findResourceById, createResource, removeResource, updateResource } = require('../util/mongo.controller');
+
+const baseCreate = async (data, resourceType) => {
+  return await createResource(data, resourceType);
+};
+const baseSearchById = async (id, resourceType) => {
+  return await findResourceById(id, resourceType);
+};
+const baseUpdate = async (id, data, resourceType) => {
+  return await updateResource(id, data, resourceType);
+};
+const baseRemove = async (id, resourceType) => {
+  return await removeResource(id, resourceType);
+};
+module.exports = { baseCreate, baseSearchById, baseUpdate, baseRemove };
--- a/src/services/patient.service.js
+++ b/src/services/patient.service.js
@@ -1,20 +1,24 @@
 const { baseCreate, baseSearchById, baseRemove, baseUpdate } = require('./base.service');
 
-const create = async (args, { req }) => {
-  return await baseCreate(req.body, 'Patient');
+//send POST request to {BASE_URL}/4_0_0/patient
+//body of request will include data to push to patient
+const create = async (_, data) => {
+  return await baseCreate(data, 'Patient');
 };
 //send GET request to {BASE_URL}/4_0_0/patient/{id}
-const searchById = async (args, { req }) => {
-  return await baseSearchById(args.id, 'Patient');
+const searchById = async args => {
+  return await baseSearchById(args, 'Patient');
 };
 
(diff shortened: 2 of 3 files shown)


mgramigna:
This PR doesn't run. I think the path in `index.js` for pointing to the `patient.service.js` file needs to be updated. It should be `./src/services/patient.service.js` now

mgramigna (changes requested):
[MG1] Upserting logic doesn't quite work, as we get an error returned. The database operation still works, but the return result throws an error:
image

This is because the `results` variable in the `updateResource` function is different when an insert happens vs. an update. Updating an existing resource works just fine, but using the PUT request to create a new resource doesn't because `results.value` is null, so we can't access `id` on it.

To replicate this behavior, send a PUT request for a brand new resource to `http://localhost:3000/4_0_0/Patient/some-id`, and you'll see that error. See what you can do to get a proper return result when inserting via the PUT request happens rather than updating

[MG2] We should sync up the `_id` and the `id` properties of the resources that are created, since `id` should be the unique identifier indexed by mongo. If we don't do this, we're allowed to create infinite number of resources with the same id. This is the current behavior of this branch. During any creation logic (regular POST and upsert), if `id` is specified, we should se `_id` to that. If it isn't, we should either generate a UUID using node, or take the ObjectID generated by mongo and set that as the ID. I think the former will probably be easier.

[MG3] We can use the `resolveSchema` function in asymmetrik to get rid of `_` properties on the return objects during `searchById`. Notice that if you do the GET request, the `_id` from mongo will be there. Using the schema pattern can clean out the return result for us.

[MGX] Inline comments

natjoe4:
add a check if the request body is blank to tell them to change the headers

mgramigna:
Regarding IDs:

* In the case of POST requests (create), the server should always be generating UUIDs, never persisting the one that's in the request body. This is an easy fix.
* In the case of PUT requests that create a new resource (when it doesn't exist in the DB before), we need to enforce that the ID is always there and matches

mgramigna:
> add a check if the request body is blank to tell them to change the headers

I think in this case, we should also check if the header isn't right. It's possible they provide an empty body outright but with the correct header, which is a different kind of error haha

PR #6, opened 2021-08-31
Transaction bundle upload

sarahmcdougall:
# Summary
The server now supports transaction Bundle uploads.

## New behavior
Transaction bundles can now be POSTed to the `deqm-test-server` such that all of the resources in the bundle are added to the database.

## Code changes
The file `src/services/bundle.service.js` was created to support the uploading of transaction bundles to the server. The `uploadTransactionBundle()` function parses out the entries from the request body and creates the `destinationUrl` for each entry, then uses `axios` to make the PUT/POST request for each entry. These requests are then passed into `makeTransactionResponseBundle()` with the type set to `’transaction-response’`, which creates a transaction-response bundle. 

The file `src/services/bundle.controller.js` was created to establish a transaction controller that executes `uploadTransactionBundle()`. 

The file `src/server/server.js` was created to establish a `DEQMServer` class that extends from the `Server` class and a function `initialize()` to create and configure a new `DEQMServer` instance. This is used in `src/index.js` to build the config and initialize the server. 

# Testing guidance
* First, get MongoDB running by using the following command: `brew services start mongodb-community@5.0`. You can verify that MongoDB is running by running the command `brew services list`. 
* Run `node src/index.js` from the terminal.
* In Insomnia, run the following tests:

1. Submit a POST request with `simple-bundle-post-entry.json.js` (in Slack). Check that you receive a `”201 Created”` status, and that the location is `”4_0_0/Patient/{id}”`. Then submit a GET request using the location id: `http://localhost:3000/4_0_0/Patient/{id}`.
2. Submit a POST request with `bundle-3-resources.json.js` (in Slack). This has a bundle with different resources of different types that reference each other. When you submit the request, you should see three entries, one of type `Encounter`, one of type `Patient`, and one of type `Procedure`, all with different locations. Note that you will get a `”200 OK”` status instead of `”201 Created”` because it has PUT requests. Again, try to submit a GET request using the locations for these entries.
3. Try to reach the two `ServerError`s in `src/services/bundle.service.js` by uploading a request that is not of `resourceType === ‘Bundle’`, and by uploading a request that is not of `type === ‘transaction’`.


hossenlopp (changes requested):
This also cannot handle POSTing a measure bundle such as EXM130-7.3.000. An error is returned "request entity too large". This may be a solved game for our express configuration in `fqm-execution-service`.

src/index.js
```diff
@@ -1,9 +1,15 @@
-const { initialize, loggers } = require('@asymmetrik/node-fhir-server-core');
+const { loggers } = require('@asymmetrik/node-fhir-server-core');
 const mongoUtil = require('./util/mongo');
 const { buildConfig } = require('./util/config');
+const { initialize } = require('./server/server');
+const express = require('express');
+const bodyParser = require('body-parser');
+
+const app = express();
+app.use(bodyParser.json({ limit: '50mb', type: 'application/json+fhir' }));
```
hossenlopp, line 9: So it turns out `body-parser' is actually included in express now and can be used directly. This should solve the deprecation warning. Also `body-parser` doesn't need to be included in the package.json.
```suggestion
const express = require('express');

const app = express();
app.use(express.json({ limit: '50mb', type: 'application/json+fhir' }));
```

hossenlopp (changes requested):
Quick fix for the `bodyParsers` deprecation.

hossenlopp (approved):
This is a good and very functional first pass. In the future we should try to make it call the code processing the bundle entries without going though axios/http request. There may be a somewhat hidden API in express we can use to do this.

PR #7, opened 2021-09-02
Submit data

natjoe4:
# Summary
We can now use custom operations to perform submit-data requests to the server

## New behavior
The server can interpret a FHIR Parameters object passed in using a POST request to http://localhost:3000/4_0_0/Measure/$submit-data. The server will parse the parameters and add new documents to the database for each contained FHIR object.

## Code changes
 - Added a transactionBundle Class to provide easy conversion from FHIR parameters -> transactionBundle -> JSON
 - Added new custom submitData operation for Measure type in config
 - Added submitData function which parses Parameters object into a transaction bundle and calls uploadTransactionBundle

# Testing guidance
- Start up mongodb using `brew services start mongodb-community@5.0` then run `npm run db:reset` and finally, `npm start`
- Open Robomongo and ensure it is hooked into the live database
- Create a valid FHIR parameters object, or use an existing one. I used the object found on [this page](https://www.hl7.org/fhir/measure-operation-submit-data.html) 
- Paste the object into the body of a request using Insomnia. 
- Submit the body in a POST request to `http://localhost:3000/4_0_0/Measure/$submit-data`
- You should get a 200 ok response with a list of entries and locations to find the created resources
- Use Robomongo to ensure that all resources in the parameters have been created
- Submit GET requests to `http://localhost:3000/{location url} for each response location url and ensure all created resources are GETtable
- Update and edit the Parameters object to try to break the upload, and ensure the server's response is appropriate


src/services/measure.service.js
```diff
@@ -42,4 +45,72 @@ const remove = async args => {
...
+    });
+  }
+
+  const tb = new TransactionBundle();
+  const resources = req.body.parameter;
+
+  let containsMeasureReport = false;
+
+  resources.forEach(resource => {
+    //TODOMAYBE: add functionality for if resource is itself a bundle
```
natjoe4, line 91: Probably will need to handle the case where a measure ID is passed in as an arg here. What is the intended behavior if this happens?

src/util/config.js
```diff
@@ -30,7 +30,15 @@ const buildConfig = () => {
...
           service: path.resolve('src', 'services', 'measure.service.js'),
-          versions: [VERSIONS['4_0_0']]
+          versions: [VERSIONS['4_0_0']],
+          operation: [
+            {
+              name: 'submitData',
+              route: '/$submit-data',
+              method: 'POST',
+              reference: 'http://hl7.org/fhir/OperationDefinition/Measure-submit-data'
+            }
```
natjoe4, line 40: should add another custom operation that allows a measure ID to be passed in as an arg in the URL

src/services/measure.service.js
```diff
@@ -42,4 +45,72 @@ const remove = async args => {
...
+          code: 'BadRequest',
+          details: {
+            text: `Expected at least one resource with name: 'measureReport' and resourceType: 'MeasureReport.`
+          }
+        }
+      ]
+    });
+  }
+  req.body = tb.toJSON();
+  const output = await uploadTransactionBundle(req, req.res);
```
natjoe4, line 112: I don't totally understand the res attribute here, so @reviewer take a closer look here

src/services/measure.service.js
```diff
@@ -42,4 +45,72 @@ const remove = async args => {
...
+const submitData = async (args, { req }) => {
+  if (req.body.resourceType !== 'Parameters') {
+    throw new ServerError(null, {
+      statusCode: 400,
+      issue: [
+        {
+          severity: 'error',
+          code: 'BadRequest',
+          details: {
+            text: `Expected 'resourceType: Parameters'. Received 'type: ${req.resourceType}'.`
```
sarahmcdougall, line 64:
```suggestion
            text: `Expected 'resourceType: Parameters'. Received 'type: ${req.body.resourceType}'.`
```

src/services/measure.service.js
```diff
@@ -42,4 +45,72 @@ const remove = async args => {
...
+      ]
+    });
+  }
+
+  const tb = new TransactionBundle();
+  const resources = req.body.parameter;
+
+  let containsMeasureReport = false;
+
+  resources.forEach(resource => {
```
sarahmcdougall, line 90: Maybe change the name of `resource` here so that you don't have to do `resource.resource` later on in the code with `tb.addEntryFromResources()`. It seems a little confusing to me to have `resource.resource`.

src/services/measure.service.js
```diff
@@ -42,4 +45,72 @@ const remove = async args => {
...
+    });
+  }
+
+  const tb = new TransactionBundle();
+  const parameters = req.body.parameter;
+
+  let containsMeasureReport = false;
+
+  parameters.forEach(param => {
+    //TODOMAYBE: add functionality for if resource is itself a bundle
```
sarahmcdougall, line 91: I agree that functionality should be added if the resource is a bundle, and it looks like `cqf-ruler` accounts for this. Should this be added as a separate task?

src/services/measure.service.js
sarahmcdougall: With http requests, the `res` attribute refers to the response body. Asymmetrik uses Express under the hood, and the functions that handle the http requests take in both a request (`req`) and a response (`res`). I didn't know that the `res` attribute can belong to `req`, but when doing a `console.log` statement it looks like `req.res` does represent the response object that we need for `uploadTransactionBundle()`.

sarahmcdougall (approved):
Everything functions as it should, and I was able to test the ServerErrors and they seem good. It may be a good idea to add functionality in `submitData` for if the resource if a bundle, so we can consider making a separate task for that (ask Matt). 

Like we talked about, a measure ID can be passed in as an arg, but I’m also not sure why this would be desired and whether it should be implemented as a custom operation. I am leaning toward yes for creating a custom operation for this (perhaps in a separate task), but would appreciate input from a second reviewer.

src/util/config.js
mgramigna: Yes. Per the [spec](https://www.hl7.org/fhir/measure-operation-submit-data.html), it should be supported on both `Measure/$submit-data` and `Measure/[id]/$submit-data`

Hopefully, the same function will invoke in both cases, the args passed into the function will just be different? Worth checking.

src/services/measure.service.js
mgramigna: Unsure what the difference in behavior is.. cqf-ruler only supports the `Measure/id/$submit-data` version. I don't think we need to worry about different functionality yet. We can bring it up at the conenctathon

src/services/measure.service.js
```diff
@@ -42,4 +45,72 @@ const remove = async args => {
...
+          }
+        }
+      ]
+    });
+  }
+
+  const tb = new TransactionBundle();
+  const parameters = req.body.parameter;
+
+  let containsMeasureReport = false;
```
mgramigna, line 88: I slightly prefer doing a short-circuit here rather than always going through the whole loop just to fail after if a `measureReport` parameter isn't there.

Suggested approach would be to do a `.filter` on the `parameters` resource for something where `param.name === 'measureReport'` and throw an error if that resulting array's length is `!== 1`. This will also properly check the case where there are more than 1 `measureReport` param, which is not allowed by the operation. The current behavior only checks if one doesn't exist at all.

src/transactionBundle.js
```diff
@@ -0,0 +1,30 @@
...
+  constructor() {
+    this.entry = [];
+  }
+
+  addEntryFromResource(resource) {
+    const newEntry = {
+      resource: resource,
+      request: {
+        method: 'POST',
+        url: resource.resourceType
```
mgramigna, line 15: I think we should probably avoid hardcoding the entry request method to `POST`. This is always the behavior for submit-data, but if this class gets re-used for any other operations, we might want to do the `PUT` request? We could add another argument for the method to the function body, and handle the following cases:

* If it's `POST`, keep what you have
* If it's `PUT`, the `url` should become `` `${resource.resourceType}/${resource.id}` `` (if `resource.id` doesn't exist, existing `PUT` logic should already handle that error?)

mgramigna (changes requested):
Functionality looks solid. A few comments

src/resources/transactionBundle.js
```diff
@@ -0,0 +1,51 @@
+/**
+ * Defines the TransactionBundle class which will allow us to add resources to
+ * a txn bundle in the proper format and convert the representation to JSON
+ */
+
+const { ServerError, resolveSchema } = require('@asymmetrik/node-fhir-server-core');
+
+const createTransactionBundleClass = baseVersion => {
+  const Bundle = resolveSchema(baseVersion, 'bundle');
```
natjoe4, line 9: not sure if this is necessary or if we're only focussed on 4_0_0.

mgramigna (approved):
looks great

PR #11, opened 2021-09-10
Add data requirements op and measure bundle assembly

mgramigna:
# Summary

Updates to the functionality introduced in #10, but on a clean branch with added support for $data-requirements

## New behavior

Adds the `$data-requirements` operation

## Code changes

* Update code to assemble measure bundle from measure ID
* Add custom op for data requirements that calls `fqm-execution`'s `$data-requirements` function
* Update dependencies with lodash and fqm-e

# Testing guidance

POST a measure bundle from the `connectathon` repo into your database using regular transaction bundle upload. Note the measure ID of the created Measure resource (e.g. `measure-EXM130-7.3.000`).

Send GET request to `http://localhost:3000/4_0_0/Measure/<your-measure-id>/$data-requirements`

Output should be a FHIR Library `module-definition` with all the data requirements

### DIFF
--- a/src/util/measureBundle.js
+++ b/src/util/measureBundle.js
@@ -17,6 +17,25 @@ function isValidLibraryUrl(s) {
   return urlRegex.test(s);
 }
 
+function hasNoDependencies(lib) {
+  return !lib.relatedArtifact || (Array.isArray(lib.relatedArtifact) && lib.relatedArtifact.length === 0);
+}
+
+function getQueryFromReference(reference) {
+  // Library references could be canonical or resourceType/id
+  if (isValidLibraryUrl(reference)) {
+    if (reference.includes('|')) {
+      const [urlPart, versionPart] = reference.split('|');
+      return { url: urlPart, version: versionPart };
+    } else {
+      return { url: reference };
+    }
+  } else {
+    const id = reference.split('/')[1];
+    return { id };
+  }
+}
+
 /**
  * Assemble a measure bundle with necessary FHIR Library resources
  * @param {string} measureId id of the measure to assemble bundle for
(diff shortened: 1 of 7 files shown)


src/util/measureBundle.js
```diff
@@ -0,0 +1,111 @@
...
+
+  // This filter checks for the 'Library' keyword on all related artifacts
+  // TODO: This filter can probably be improved, but will work in our cases for now
+  const depLibUrls = lib.relatedArtifact
+    .filter(ra => ra.type === 'depends-on' && ra.resource.includes('Library'))
+    .map(ra => ra.resource);
+
+  // Obtain all libraries referenced in the related artifact, and recurse on their dependencies
+  const libraryGets = depLibUrls.map(async url => {
+    const [urlPart, versionPart] = url.split('|');
```
hossenlopp, line 96: This doesn't handle the situation where no version is specified on the URL.

src/util/measureBundle.js
```diff
@@ -0,0 +1,111 @@
...
+      ]
+    });
+  }
+
+  const [mainLibraryRef] = measure.library;
+
+  // Library references could be canonical or resourceType/id
+  let mainLib;
+  if (isValidLibraryUrl(mainLibraryRef)) {
+    mainLib = await findOneResourceWithQuery({ url: mainLibraryRef }, 'Library');
```
hossenlopp, line 48: This doesn't handle the situation where the canonical url includes the version.

src/util/measureBundle.js
```diff
@@ -0,0 +1,111 @@
...
+  const libraryGets = depLibUrls.map(async url => {
+    const [urlPart, versionPart] = url.split('|');
+
+    const lib = await findOneResourceWithQuery({ url: urlPart, version: versionPart }, 'Library');
+    return getAllDependentLibraries(lib);
+  });
+
+  const allDeps = await Promise.all(libraryGets);
+
+  results.push(...allDeps);
```
hossenlopp, line 104: There could be the possibility that this grabs duplicate copies of libraries if multiple supporting libraries use the same libraries.

hossenlopp (changes requested):
`$data-requirements` call works but there could be some issues with finding supporting libraries. The bundles for each measure do not have the ValueSets which will be needed for subsequent execution/gaps work.

src/util/measureBundle.js
mgramigna: Good catch, added a function to support this

src/util/measureBundle.js
mgramigna: Good catch, added a function to support this

src/util/measureBundle.js
mgramigna: Yeah.. the duplicate logic is not ideal. In the interest of time, we decided to leave it. I noted a TODO in there to improve the recursion to filter out already visited resources.

hossenlopp (approved):
Looks good to me. Now do execution and care gaps. 🤣

PR #13, opened 2021-09-17
Add script for parsing model info and outputted json

mgramigna:
# Summary

This PR adds a repeatable script to parse FHIR ModelInfo to identify all the ways that a FHIR resource can reference a Patient resource. The outputted JSON format is:

```JSON
{
  "<resourceType>": ["<key1 that can be used to reference patient>", "<key2 that can be used to reference patient>"]
}
```

I also filter out a hardcoded list of keys that we want to ignore. Right now, `"where(resolve() is Patient)"` is ignored. Everything else seems fine.

## New behavior

`npm run parse-references` is an available script now that will parse the model info XML and output the JSON of the above format to `src/model-info/patient-references.json`

## Code changes

* Set proper node environment in `.eslintrc` (stops eslint from complaining about built-in node global variables e.g. `__dirname`)
* Add `xml2js` as explicit dependency
* Add fhir 4.0.1 model info XML to `src/model-info`
* Add script to parse XML into outputted JSON
* Added copy of the output of the script to `src/model-info/patient-references.json`

# Testing guidance

Run the `npm run parse-references` script and inspect the output. Feel free to ensure that if aligns with the XML correctly for the `contextRelationship`s.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "start": "node src/index.js",
     "db:setup": "node src/scripts/createCollections.js",
     "db:delete": "node src/scripts/deleteCollections.js",
-    "db:reset": "npm run db:delete && npm run db:setup"
+    "db:reset": "npm run db:delete && npm run db:setup",
+    "parse-references": "node src/scripts/parseModelInfo.js"
   },
   "author": "",
   "license": "Apache-2.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -13,7 +13,8 @@
     "start": "node src/index.js",
     "db:setup": "node src/scripts/createCollections.js",
     "db:delete": "node src/scripts/deleteCollections.js",
-    "db:reset": "npm run db:delete && npm run db:setup"
+    "db:reset": "npm run db:delete && npm run db:setup",
+    "parse-references": "node src/scripts/parseModelInfo.js"
```
hossenlopp, line 17: Would suggest expanding this name later if we also have it parse data to help with search functionality. This can be done later when that work is necessary.

hossenlopp (approved):

PR #14, opened 2021-09-22
Unit testing

MegRiley:
# Summary
Added basic unit testing functionality for the measure resource
## New behavior

- added the ability have the mongo controller should connect to either a test database or the regular database. 
- added basic unit tests for crud operations using the measure resource

## Code changes

- **dbconfig.js/,env.test/mongo.js** - added a check that depends on the `process.env.NODE_ENV` either the regular database will be used or the test database will be utilized. 
- **measure.service.test.js** - added a new file that contains unit tests for the basic crud operations using a measure as the resourceType.
- **testMeasure.json/testPatient.json/testLibary.json**  - these are very basic files that contain the bare minimum data elements for each resource named that can be loaded into the database and used for unit testing

-  **populateTestData.js** - adds the functionality to load the  test database with any resource types to be used in unit testing
- **package.json** - added supertest library and expanded the jest command

# Testing guidance
The unit tests can be executed by running: `npm run test`.
 It will also push the 3 resource files into the database and remove them after the tests. (I'm not sure if there is a way to watch the database closely enough to see this, though to check the behavior you could just remove the `afterAll` call from `measure.service.test.js` and that would show the results of the insertions


test/measure.service.test.js
```diff
@@ -0,0 +1,127 @@
...
+describe('measure.service', () => {
+  beforeAll(async () => {
+    await testSetup(testMeasure, testPatient, testLibrary);
+  });
+  afterAll(async () => {
+    await cleanUpDb();
+  });
+  test('test', async () => {
+    await supertest(server.app).get('/4_0_0/metadata').expect(200);
+  });
```
mgramigna, line 63:
```suggestion
  test('test', async () => {
    await supertest(server.app).get('/4_0_0/metadata').expect(200);
  });
```

This can be removed. It was just to make sure the supertest syntax was correct IIRC, but we don't need it anymore

test/measure.service.test.js
```diff
@@ -0,0 +1,127 @@
...
+  describe('create', () => {
+    test('test create with correct headers', async () => {
+      await supertest(server.app)
+        .post('/4_0_0/Measure')
+        .send(postRequest)
+        .expect(200)
+        .then(async response => {
+          // Check the response
+          expect(response.body._id != null);
```
mgramigna, line 72: 2 things

1) this checks for the presence of a mongo-generated `_id`, which isn't bad, but probably what makes more sense to do with a POST request is to make sure the server generated a new ID, and that the content location header of the response indicates where to find it. So we should be accessing response.body.id, and making sure it was modified to some uuid that isn't the ID given by the test fixture since that is the expected behavior. You may also want to assert something about the Content-Location header.

2) For `jest` assertions, there is a [convenience function](https://jestjs.io/docs/expect#tobenull) for null assertions, and you can chain it with a `.not` for negation: `expect(something).not.toBeNull()`

test/measure.service.test.js
```diff
@@ -0,0 +1,127 @@
...
+      await supertest(server.app)
+        .get('/4_0_0/Measure/testMeasure')
+        .send(getRequest)
+        .expect(200)
+        .then(async response => {
+          // Check the response
+
+          expect(response.body.title).toBe(getRequest.title);
+          expect(response.body.content).toBe(getRequest.content);
+        });
```
mgramigna, line 90: what does this content/title assertion do?

test/measure.service.test.js
```diff
@@ -0,0 +1,127 @@
...
+    });
+  });
+  describe('update', () => {
+    //*a put request*/
+    test('test update with correctHeaders and  the id is in database', () => {
```
mgramigna, line 95: Is there a reason this doesn't use await but the others do?

test/measure.service.test.js
```diff
@@ -0,0 +1,127 @@
...
+
+          // Check the response
+          expect(response.body._id).toBeTruthy();
+          expect(response.body.title).toBe(data.title);
+          expect(response.body.content).toBe(data.content);
+        });
+    });
+  });
+  describe('remove', () => {
+    test('removing the measure from the database when the measure is indeed present', () => {
```
mgramigna, line 114: Is there a reason this doesn't use await but the others do?

test/populateTestData.js
```diff
@@ -0,0 +1,19 @@
...
+  await collection.insertOne(data);
+  return { id: data.id };
+};
+//clean up db after test
+async function cleanUpDb() {
+  await db.dropDatabase();
+  await client.close();
+}
+
+const testSetup = async (pathToFiles, pathToPatients, pathToTestLibrary) => {
```
mgramigna, line 13: Can we rename these arguments now that you updated it to pass in the json content rather than a path?

src/util/mongo.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable no-undef */
 const { MongoClient } = require('mongodb');
-
-require('dotenv').config();
+/*eslint-disable-next-line no-unused-vars*/
+const dbconfig = require('./dbconfig');
```
mgramigna, line 4:
```suggestion
require('./dbconfig');
```

package.json
```diff
@@ -28,8 +31,11 @@
     "uuid": "^8.3.2"
   },
   "devDependencies": {
+    "@shelf/jest-mongodb": "^2.0.3",
```
mgramigna, line 34: Do you need this dependency? It seems unused since we populate the test database and everything ourselves

mgramigna (changes requested):
[MG1] The tests fail for me. I think some of the supertest config is wrong. We do expect 200 on everything, but all the requests should have different response codes depending on what they're doing.

[MG2] This only tests the measure service requests, but remember that we also want tests for the base.service functions which will in effect test all other resource types that we support rn. The tests might be largely the same, but we should still be testing both.

[MG3] The definition of done for this task says to put the testing scripts the GH actions yml file. Can we add the testing script to github actions?

[MG4-X] Made some inline comments

test/measure.service.test.js
MegRiley: No reason, I had added the awaits in after I had done the first pass and missed it, I added them in

test/measure.service.test.js
MegRiley: No reason, I had added the awaits in after I had done the first pass and missed it, I added them in

test/populateTestData.js
MegRiley: I will make that change

package.json
MegRiley: removed it, I had been investigating using that to populate the database

src/util/mongo.js
MegRiley: Fixed

natjoe4 (changes requested):
I think I found where your unit tests are breaking. the content-type: application/fhir+json header isn't making it in the way its supposed to which results in an empty request body. If you add another .set() to the update test, where you set the 'content-type' header correctly, the server responds with a 200. I've also noticed that super-test seems to send whatever is passed into the .send function as the body of the request, so the info stored inside putRequest.body actually ends up being stored inside request.body.body by the time it gets to the server which breaks our code. I moved the putRequest.body attributes into the putRequest object at the top level. Additionally, the id field in the request body needs to match the id in the url for the PUT request or the request is rejected outright, so I changed this to 'testMeasure' in the putRequest object. Finally, the response.body._id seems to always come back as undefined, even if the update works. I think this is because the update function always returns an empty body upon success, but it also returns a location header with the url to send future requests to. I had success with comparing the response.headers.location to a value to ensure the request succeeded.

mgramigna (changes requested):
* I think [this comment](https://github.com/projecttacoma/deqm-test-server/pull/14#discussion_r716739411) is still outstanding, and is present in both the base service tests and the measure service tests

* I'm a little confused by the `getRequest`, `postRequest`, etc. objects that are passed into supertest. Do we even need these? We created test fixtures specifically to act as request bodies where necessary (PUT, POST). And for GET/DELETE, we don't need request bodies in the first place. Take the following example, I think this is the kind of structure we should be after for a POST request test for example:

``` JavaScript
      await supertest(server.app)
        .post('/4_0_0/Patient')
        .set('Accept', 'application/json+fhir')
        .set('content-type', 'application/json+fhir')
        .send(testPatient)
        .expect(201)
        .then(async response => {
          // do something
        });
```

This makes all the header config/response code expect happen in supertest, and then the request body is sent with `.send`, just like we do in [fqm-execution-service tests](https://github.com/projecttacoma/fqm-execution-service/blob/cb6c2a6f4bcd4fa96d49a5599dc823a84f98197e/test/app.test.ts#L23)

The following example is what I'd expect from a GET request test:

``` JavaScript
      await supertest(server.app)
        .get('/4_0_0/Patient/testPatient')
        .set('Accept', 'application/json+fhir')
        .expect(200)
        .then(async response => {
          // something
        });
```

See how there is no `.send` used at all? GET requests don't need a request body, so I don't think we need that object that you've created in the services to send along with the requests. I could be missing something here but I think this is right.

mgramigna:
Also, this branch needs a rebase. It's just changes in the package.json for a jest dependency and test commands

mgramigna:
One thing I noticed by running `npm test` multiple times is that it isn't idempotent (sometimes they all pass, sometimes they fail). 

I did some digging, and I believe this is because there are race conditions with doing reads/writes to the same db. Jest does things concurrently by default, so sometimes the measure tests will finish and drop the database before the base service tests are able to finish.

I _think_ a way around this is to add the `--runInBand` flag to the jest commands in package.json. Doing that allowed the tests to always pass for me without the concurrency issues. Whoops!

https://jestjs.io/docs/cli#--runinband

test/base.service.test.js
```diff
@@ -0,0 +1,77 @@
...
+        .post('/4_0_0/Patient')
+        .send(testPatient)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(201)
+        .then(async response => {
+          // Check the response
+          expect(response.body._id).not.toBeNull();
+          expect(response.body._id).not.toBe(testMeasure.id);
+          expect(response.headers.location).not.toBeNull();
```
mgramigna, line 29:
```suggestion
          expect(response.headers.location).toBeDefined();
```

The first 2 assertions don't really mean anything. Mongo is always gonna create a `_id` and it will never match. It's just part of mogno no need to test that with these.

Suggest making the assertion here `.toBeDefined()`, I think if it's undefined, that would still pass this assertion even though that's undesired

test/base.service.test.js
```diff
@@ -0,0 +1,77 @@
...
+        });
+    });
+  });
+  describe('searchById', () => {
+    //* result of sending a GET request to {BASE_URL}/4_0_0/Measure/{id}
```
mgramigna, line 34:
```suggestion
```

test/base.service.test.js
```diff
@@ -0,0 +1,77 @@
...
+    test('test searchById with correctHeaders and  the id should be in database', async () => {
+      await supertest(server.app)
+        .get('/4_0_0/Patient/testPatient')
+        .set('Accept', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          expect(response.statusCode).not.toBe('ResourceNotFound');
```
mgramigna, line 41:
```suggestion
          expect(response.body.id).toEqual(testPatient.id);
```

This assertion doesn't do anything. The status code is already asserted via the `.expect(200)`. We should be asserting that the retrieved resource matches the fixture that we expect to be there by comparing the IDs

test/base.service.test.js
```diff
@@ -0,0 +1,77 @@
...
+    test('removing the data from the database when the data is indeed present', async () => {
+      await supertest(server.app)
+        .delete('/4_0_0/Patient/testPatient')
+        .send(testPatient)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(204)
+        .then(async response => {
+          expect(response.statusCode).toBe(204);
+        });
```
mgramigna, line 70: This assertion is irrelevant because of the expect(204) above. This whole .then block can be deleted
```suggestion
```

test/measure.service.test.js
```diff
@@ -0,0 +1,84 @@
...
+  });
+  describe('remove', () => {
+    test('removing the measure from the database when the measure is indeed present', async () => {
+      await supertest(server.app)
+        .delete('/4_0_0/Measure/testMeasure')
+        .send(testMeasure)
+        .expect(204)
+        .then(async response => {
+          // Check the response
+          expect(response.statusCode).toBe(204);
```
mgramigna, line 76: see comment in base.service test, same here

test/measure.service.test.js
```diff
@@ -0,0 +1,84 @@
...
+        .put('/4_0_0/Measure/testMeasure')
+        .send(updateMeasure)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          // Check the response
+          expect(response.headers.location).toBeTruthy();
+          expect(response.body.location).not.toBeNull();
+          console.log(response.error.message);
```
mgramigna, line 64:
```suggestion
          expect(response.headers.location).toBeDefined();
```

test/measure.service.test.js
```diff
@@ -0,0 +1,84 @@
...
+    test('test searchById with correctHeaders and  the id should be in database', async () => {
+      await supertest(server.app)
+        .get('/4_0_0/Measure/testMeasure')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          // Check the response
+          expect(response.body._id).not.toBeNull();
+          expect(response.statusCode).not.toBe('ResourceNotFound');
```
mgramigna, line 47: Same deal as base.service. see above comment

test/measure.service.test.js
```diff
@@ -0,0 +1,84 @@
...
+      await supertest(server.app)
+        .post('/4_0_0/Measure')
+        .send(testMeasure)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(201)
+        .then(response => {
+          // Check the response
+          expect(response.body._id).not.toBeNull();
+          expect(response.headers.location).not.toBeNull();
```
mgramigna, line 28: same as base.service, see above comment

test/measure.service.test.js
```diff
@@ -0,0 +1,84 @@
...
+        })
+        .catch(error => {
+          console.log('we have error', error);
+        });
```
mgramigna, line 32:
```suggestion
```

Assuming this was just for debugging

mgramigna (changes requested):

test/measure.service.test.js
```diff
@@ -0,0 +1,66 @@
...
+
+  describe('searchById', () => {
+    test('test searchById with correctHeaders and  the id should be in database', async () => {
+      await supertest(server.app)
+        .get('/4_0_0/Measure/testMeasure')
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          expect(response.body.id).toEqual(testPatient.id);
```
mgramigna, line 39:
```suggestion
          expect(response.body.id).toEqual(testMeasure.id);
```

copy paste error causes test fail

test/base.service.test.js
```diff
@@ -0,0 +1,71 @@
...
+    test('test update with correctHeaders and  the id is in database', async () => {
+      await supertest(server.app)
+        .put('/4_0_0/Patient/testPatient')
+        .send(updatePatient)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          // Check the response
+          expect(response.body.location).not.toBeNull();
```
natjoe4, line 53:
```suggestion
          expect(response.headers.location).not.toBeNull();
```
response.body.location is undefined which is not null, so the test passes, but I imagine we want the response.headers.location here

test/base.service.test.js
```diff
@@ -0,0 +1,71 @@
...
+        .set('content-type', 'application/json+fhir')
+        .expect(201)
+        .then(async response => {
+          // Check the response
+          expect(response.headers.location).toBeDefined();
+        });
+    });
+  });
+  describe('searchById', () => {
+    test('test searchById with correctHeaders and  the id should be in database', async () => {
```
natjoe4, line 32:
```suggestion
    test('test searchById with correctHeaders and the id should be in database', async () => {
```
nit pick

test/base.service.test.js
```diff
@@ -0,0 +1,71 @@
...
+        .set('Accept', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          expect(response.body.id).toEqual(testPatient.id);
+        });
+    });
+  });
+  describe('update', () => {
+    //*a put request*/
+    test('test update with correctHeaders and  the id is in database', async () => {
```
natjoe4, line 44:
```suggestion
    test('test update with correctHeaders and the id is in database', async () => {
```
same here

test/base.service.test.js
```diff
@@ -0,0 +1,71 @@
...
+      await supertest(server.app)
+        .put('/4_0_0/Patient/testPatient')
+        .send(updatePatient)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          // Check the response
+          expect(response.headers.location).not.toBeNull();
```
natjoe4, line 53:
```suggestion
          expect(response.headers.location).toBeDefined();
```
sorry should have included this in last comment!! I think it makes more sense if this matches create test.

package.json
```diff
@@ -4,19 +4,22 @@
...
     "start": "node src/index.js",
     "start:watch": "nodemon src/index.js",
     "db:setup": "node src/scripts/createCollections.js",
     "db:delete": "node src/scripts/deleteCollections.js",
     "db:reset": "npm run db:delete && npm run db:setup",
     "parse-references": "node src/scripts/parseModelInfo.js",
-    "connectathon-upload": "node src/scripts/getConnectathonBundles.js"
+    "connectathon-upload": "node src/scripts/getConnectathonBundles.js",
+    "test": "jest --silent --runInBand",
+    "test:coverage": "jest --collectCoverage --silent",
```
mgramigna, line 20:
```suggestion
    "test:coverage": "jest --collectCoverage --silent --runInBand",
```

test/measure.service.test.js
```diff
@@ -0,0 +1,66 @@
...
+        .set('content-type', 'application/json+fhir')
+        .expect(201)
+        .then(response => {
+          expect(response.headers.location).toBeDefined();
+        });
+    });
+  });
+
+  describe('searchById', () => {
+    test('test searchById with correctHeaders and  the id should be in database', async () => {
```
natjoe4, line 32:
```suggestion
    test('test searchById with correctHeaders and the id should be in database', async () => {
```

test/measure.service.test.js
```diff
@@ -0,0 +1,66 @@
...
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .expect(200)
+        .then(async response => {
+          expect(response.body.id).toEqual(testMeasure.id);
+        });
+    });
+  });
+  describe('update', () => {
+    test('test update with correctHeaders and  the id is in database', async () => {
```
natjoe4, line 44:
```suggestion
    test('test update with correctHeaders and the id is in database', async () => {
```

test/measure.service.test.js
```diff
@@ -0,0 +1,66 @@
...
+        .expect(200)
+        .then(async response => {
+          // Check the response
+          expect(response.headers.location).toBeDefined();
+        });
+    });
+  });
+  describe('remove', () => {
+    test('removing the measure from the database when the measure is indeed present', async () => {
+      await supertest(server.app).delete('/4_0_0/Measure/testMeasure').send(testMeasure).expect(204);
```
natjoe4, line 59:
```suggestion
      await supertest(server.app).delete('/4_0_0/Measure/testMeasure').expect(204);
```
We shouldn't need a request body for delete requests

mgramigna (approved):

natjoe4 (approved):

PR #15, opened 2021-09-22
Base Search Implementation

hossenlopp:
# Summary
Base search implementation that uses Asymmetrik's fhir-qb to build the mongo queries. 

## New behavior
Implements basic search parameters. `_id`, `_lastUpdated`, `_tag`, `_profile`, `_security`, `identifier` for all resources. Provides a way for services to use the base search functionality with resource specific parameter definitions. Measure service uses this to provide searching on `name` and `version`.

## Code changes
- Added baseSearch implementation to the base service.
- Added mongo controller capability to run aggregation query.
- Replaced Measure service search with a call to use the baseSearch with additional parameter definitions provided.

# Testing guidance
Load EXM130 bundle into the server. Run searching queries from the testing guidance in #12. 

Make sure some of the following url's return results in the search set.
```
http://localhost:3000/4_0_0/Patient?_profile=http%3A%2F%2Fhl7.org%2Ffhir%2Fus%2Fcore%2FStructureDefinition%2Fus-core-patient
```
```
http://localhost:3000/4_0_0/Measure?_id=measure-EXM130-7.3.000
```
```
http://localhost:3000/4_0_0/Patient
```

Make sure these don't return results.
```
http://localhost:3000/4_0_0/Measure?_id=measure-EXM69-4.2.000
```
```
http://localhost:3000/4_0_0/Measure?name=EXM000
```

Test with other parameters that are not supported (ex. `publisher`). You should get 400 errors stating so.

### DIFF
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -28,7 +28,10 @@ const globalParameterDefinitions = {
   }
 };
 
-const qb = new QueryBuilder({ globalParameterDefinitions });
+const qb = new QueryBuilder({
+  globalParameterDefinitions,
+  implementationParameters: { archivedParamPath: '_isArchived' }
+});
 
 /**
  * creates an object and generates an id for it regardless of the id passed in
@@ -73,13 +76,15 @@ const baseSearchById = async (args, resourceType) => {
 };
 
 const baseSearch = async (args, context, resourceType) => {
-  console.log(context.req);
+  //console.log(context.req);
   const dataType = resolveSchema(args.base_version, resourceType.toLowerCase());
   const Bundle = resolveSchema(args.base_version, 'bundle');
   const BundleEntry = resolveSchema(args.base_version, 'bundleentry');
+  context.req.params = {};
   const filter = qb.buildSearchQuery(context);
-  console.log(filter);
+  console.log(JSON.stringify(filter));
   const results = await (await findResourcesWithFilter(filter, resourceType)).toArray();
+  console.log(results);
   const resultResources = results.map(result => new dataType(result));
   const entries = resultResources.map(resource => new BundleEntry({ resource: resource }));
   return new Bundle({ entry: entries });
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -1,13 +1,34 @@
 const { v4: uuidv4 } = require('uuid');
 const { resolveSchema, ServerError } = require('@asymmetrik/node-fhir-server-core');
 const {
-  findResourcesWithQuery,
   findResourceById,
   createResource,
   removeResource,
-  updateResource
+  updateResource,
+  findResourcesWithFilter
 } = require('../util/mongo.controller');
-const { mapArrayToSearchSetBundle } = require('../util/bundleUtils');
+const QueryBuilder = require('@asymmetrik/fhir-qb');
+
+const globalParameterDefinitions = {
+  _content: {
+    type: 'string',
+    fhirtype: 'string',
+    xpath: '',
+    definition: 'http://hl7.org/fhir/SearchParameter/Resource-content',
+    description: 'Search on the entire content of the resource',
+    modifier: 'missing,exact,contains'
+  },
+  _id: {
+    type: 'token',
+    fhirtype: 'token',
+    xpath: 'Resource.id',
+    definition: 'http://hl7.org/fhir/SearchParameter/Resource-id',
+    description: 'Logical id of this artifact',
+    modifier: 'missing,text,not,in,not-in,below,above,ofType'
+  }
+};
+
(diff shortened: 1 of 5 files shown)


src/services/base.service.js
```diff
@@ -51,6 +108,73 @@ const baseSearchById = async (args, resourceType) => {
...
+  const Bundle = resolveSchema(args.base_version, 'bundle');
+
+  // wipe out params since the 'base_version' here breaks the query building
+  req.params = {};
+
+  // build result bundle. default to an empty result
+  const searchBundle = new Bundle({
+    type: 'searchset',
+    meta: { lastUpdated: new Date().toISOString() },
+    total: 0
```
mgramigna, line 132: NIT: I couldn't find anything in the FHIR spec for whether or not this matters, but maybe it's better consistency to have a default of `entry: []` instead of omitting is so that the empty bundles just have an empty entry array rather than a non-existent one

src/services/base.service.js
```diff
@@ -51,6 +108,73 @@ const baseSearchById = async (args, resourceType) => {
...
+ * @param {*} paramDefs Optional parameter definitions for the specific resource types. Specific
+ *                      resource services should call this and pass
+ * @returns Search set result bundle
+ */
+const baseSearch = async (args, { req }, resourceType, paramDefs = {}) => {
+  // grab the schemas for the data type and Bundle to use for response
+  const dataType = resolveSchema(args.base_version, resourceType.toLowerCase());
+  const Bundle = resolveSchema(args.base_version, 'bundle');
+
+  // wipe out params since the 'base_version' here breaks the query building
```
mgramigna, line 125: oof

src/services/base.service.js
```diff
@@ -51,6 +108,73 @@ const baseSearchById = async (args, resourceType) => {
...
+    total: 0
+  });
+
+  // build the aggregation query
+  const filter = qb.buildSearchQuery({ req: req, includeArchived: true, parameterDefinitions: paramDefs });
+
+  // if the query builder was able to build a query actually execute it.
+  if (filter.query) {
+    // grab the results from aggregation. has metadata about counts and data with resources
+    const results = (await (await findResourcesWithAggregation(filter.query, resourceType)).toArray())[0];
```
mgramigna, line 141: Is there a benefit at all to pawning off the `toArray` bit to the mongo controller function instead, to make it always return an array?

This is what we did with the `findResourcesWithQuery` function, so might be worth it for consistency

src/services/base.service.js
```diff
@@ -51,6 +108,73 @@ const baseSearchById = async (args, resourceType) => {
...
+
+    // If this is undefined, there are no results.
+    if (results.metadata[0]) {
```
mgramigna, line 144: is there a case where `results` could be undefined? should we add a `results && results.metadata[0]` case instead?

Possibly not, just spitballin lol

src/util/mongo.controller.js
```diff
@@ -73,11 +73,17 @@ const removeResource = async (id, resourceType) => {
   return collection.deleteOne({ id: id });
 };
 
+const findResourcesWithAggregation = async (query, resourceType) => {
```
mgramigna, line 76: Can you add a JSDoc here explaining what doing the aggregate does for this query?

mgramigna (commented):
Code looks good and does the things! I have no real change requests, just a few comments to consider

src/services/base.service.js
MegRiley: I had a similar question but wasn't sure if it was just a really rare edge case.

MegRiley (commented):
Looks good to me

src/services/base.service.js
hossenlopp: Resolved in slack. HAPI doesn't return an empty entry array, so we wont either.

src/services/base.service.js
hossenlopp: yeah... I don't get it.

src/services/base.service.js
hossenlopp: Yeah. I can make it consistent. :)

src/services/base.service.js
hossenlopp: Good call. It is possible, but seems to only happen when the query building comes up with a goose egg which it also returns an error.

src/util/mongo.controller.js
hossenlopp: Yep! Totally missed this when commenting.

hossenlopp:
Rebased on `main` and addressed comments.

hossenlopp:
Rebased again. Ready for final re-review.

mgramigna (approved):
Looks good post rebase

MegRiley (approved):
Looks good to me

PR #19, opened 2021-09-27
Status location

natjoe4:
# Summary
Deqm-test-server now responds with a 202 status code and a content location header when sent a submit-data request with the prefer header set to respond-async.

## New behavior
When sending a submit-data request with the prefer header set to 'respond-async', a barebones bulk data import function is now called which returns a 202 status code and a (hard-coded for now) content location to query for import status updates. The user can now send follow-up get requests to that endpoint and receive a 501 NotImplemented error.

## Code changes
- Implemented a bulkImport function that returns a 202 status code and a content-location header
- Added a bulkstatus controller which manages the flow of bulk data import status requests.
- Exposed a /bulkstatus/:id endpoint which can receive requests for bulk import status updates, but for now just returns a 501 NotImplemented error. 

# Testing guidance
- check out Sarah's recent PR for information on how to send bulkImport requests [here](https://github.com/projecttacoma/deqm-test-server/pull/18).
- Send a bulkImport request and ensure that the response code is a `202` and that the headers contain a contain a `Content-Location` field which is set to `EXAMPLE-LOCATION`.
- Send a `GET` request to `http://localhost:3000/4_0_0/bulkstatus/EXAMPLE-LOCATION` and ensure you receive a 501 NotImplemented error.
- If you have input on how to avoid the hijinks I pull on line 162 of measure.service.js to get Asymmetrik not to overwrite the status code, please let me know

### DIFF
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -141,13 +141,12 @@ const submitData = async (args, { req }) => {
 };
 
 /**
- * TODO: implement bulk import stuff
+ * "TO-DO: add bulk import funtionality" (sic)
  * @param {*} args the args object passed in by the user
  * @param {*} req the request object passed in by the user
  */
 // eslint-disable-next-line no-unused-vars
 const bulkImport = async (args, { req }) => {
-  logger.info('Measure >>> $bulk-import');
   const res = req.res;
   logger.info('Measure >>> $bulk-import');
   res.status(202);
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -147,6 +147,7 @@ const submitData = async (args, { req }) => {
  */
 // eslint-disable-next-line no-unused-vars
 const bulkImport = async (args, { req }) => {
+  logger.info('Measure >>> $bulk-import');
   const res = req.res;
   logger.info('Measure >>> $bulk-import');
   res.status(202);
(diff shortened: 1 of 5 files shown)


src/services/measure.service.js
```diff
@@ -141,25 +141,21 @@ const submitData = async (args, { req }) => {
...
 
 /**
- * TO-DO: add bulk import functionality
+ * TODO: implement bulk import stuff
  * @param {*} args the args object passed in by the user
  * @param {*} req the request object passed in by the user
  */
 // eslint-disable-next-line no-unused-vars
 const bulkImport = async (args, { req }) => {
   logger.info('Measure >>> $bulk-import');
```
sarahmcdougall, line 150:
```suggestion
```

sarahmcdougall: We have this logger.info included twice

src/services/measure.service.js
```diff
@@ -141,25 +141,21 @@ const submitData = async (args, { req }) => {
 };
 
 /**
- * TO-DO: add bulk import functionality
+ * TODO: implement bulk import stuff
```
sarahmcdougall, line 144:
```suggestion
 * TO-DO: add bulk import functionality
```

sarahmcdougall (changes requested):
Good work! All the functionality works as expected. I just have a few suggested changes but nothing major. Unfortunately, I don’t really have a solution to the measure.service.js code for overwriting the status code :/ I think it suffices for now

src/services/measure.service.js
natjoe4: I'll consider it...

sarahmcdougall (approved):
Everything looks good now, aside from the sass you've been giving me. Thanks for somewhat changing the comment back

PR #21, opened 2021-09-27
Connectathon script

sarahmcdougall:
# Summary
A new script was created that inserts the measure bundles from the Connectathon repository into our MongoDB database.

## New behavior
We can now insert all the measure bundles from the Connectathon repository into our database, which we can utilize in future tasks.

## Code changes
The `package.json` file was updated so that we can now use the command `npm run connectathon-upload` to run the `getConnectathonBundles.js` script. This script makes use of a gitignored connectathon repository and retrieves all the EXM bundle `JSON` files using the `getBundleFiles` function. Then, the script connects to the database, parses through each bundle file, and uploads the bundles’ resources to the appropriate collections using the `createResource` function from the mongo controller. 

# Testing guidance
`Git clone` the `connectathon` repository into the root directory of the `deqm-test-server` repository. Then, run `npm run connectathon-upload`, which will upload all the resources from the EXM bundles into the `deqm-test-server` database. Use Robo 3T to check that the appropriate collections (Patient, Encounter, ValueSet, Condition, etc.) are now populated. Note that some of the bundles make use of the same ValueSets and Libraries, so you should run into duplicate key errors in the terminal.

### DIFF
--- a/src/scripts/getConnectathonBundles.js
+++ b/src/scripts/getConnectathonBundles.js
@@ -1,0 +1,34 @@
+const fs = require('fs');
+const path = require('path');
+
+const connectathonPath = path.resolve(path.join(__dirname, '../../connectathon/fhir401/bundles/measure/'));
+
+/**
+ * Sends POST requests to the transaction bundle endpoint for every measure in the
+ * connectathon repo.
+ *
+ * TODO: Currently configured for fhir401 connectathon measure bundles,
+ * but may want to expand to other measure bundle providers in the future.
+ */
+async function main() {
+  console.log('hey');
+  console.log(connectathonPath);
+  // first try to just parse through the repo and get the measure bundle for each bundle
+  fs.readdir(connectathonPath, function (err, subDirs) {
+    if (err) {
+      console.error('could not access directories within connectathon repo');
+    }
+
+    subDirs.forEach(dir => {
+      // this works
+      console.log(dir);
+    });
+  });
+}
+
+main()
+  .catch(console.error)
+  .catch(e => {
+    console.error(e);
+  });
+//.finally(() => mongoUtil.client.close());
--- a/src/scripts/getConnectathonBundles.js
+++ b/src/scripts/getConnectathonBundles.js
@@ -1,34 +1,68 @@
 const fs = require('fs');
 const path = require('path');
+const mongoUtil = require('../util/mongo');
+const { createResource } = require('../util/mongo.controller');
 
 const connectathonPath = path.resolve(path.join(__dirname, '../../connectathon/fhir401/bundles/measure/'));
+// bundles that are not the latest available version
+const IGNORED_BUNDLES = ['EXM347-3.2.000-bundle.json'];
+// files containing EXM bundles of interest from connectathon repo
+const bundleFiles = [];
 
 /**
- * Sends POST requests to the transaction bundle endpoint for every measure in the
- * connectathon repo.
+ * Retrieves EXM bundle files from the connectathon repo using a regular expression.
+ * Uses recursion to parse through all available subdirectories.
+ * @param {string} directory - directory path to start at
+ * @returns array of string paths that represent the bundle files of interest
+ */
+const getBundleFiles = directory => {
+  const fileNameRegExp = new RegExp(/(^EXM.*.json$)/);
+  const filesInDirectory = fs.readdirSync(directory);
+  filesInDirectory.forEach(file => {
+    const absolute = path.join(directory, file);
+    if (fs.statSync(absolute).isDirectory()) {
+      getBundleFiles(absolute);
+    } else if (fileNameRegExp.test(file) && !IGNORED_BUNDLES.includes(file)) {
+      bundleFiles.push(absolute);
+    }
+  });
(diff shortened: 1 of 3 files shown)


src/scripts/getConnectathonBundles.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Uploads all the resources from the connectathon measure bundles into the
+ * database.
+ *
+ * TODO: Currently configured for fhir401 connectathon measure bundles,
+ * but may want to expand to other measure bundle providers in the future.
+ */
+async function main() {
+  const bundleFiles = getBundleFiles(connectathonPath);
```
natjoe4, line 40: I think it's a little confusing to overwrite this global variable here. Consider removing the return statement from getBundleFiles and then just calling getBundleFiles here and accessing the global bundleFiles variable which should be the same.

src/scripts/getConnectathonBundles.js
```diff
@@ -0,0 +1,68 @@
...
+
+  await mongoUtil.client.connect();
+  console.log('Connected successfully to server');
+
+  const bundlePromises = bundleFiles.map(async filePath => {
+    // read each EXM bundle file
+    const data = fs.readFileSync(filePath, 'utf8');
+    if (data) {
+      const bundle = await JSON.parse(data);
```
natjoe4, line 49:
```suggestion
      const bundle = JSON.parse(data);
```
JSON.parse is synchronous so this await is unnecessary

natjoe4 (changes requested):
This looks good to me and the functionality was as advertised. I left a couple of very small comments. I'm still pretty shaky on async logic inside map functions, so I'd recommend having one more person look that over if you haven't already. Good work!

src/scripts/getConnectathonBundles.js
sarahmcdougall: Good catch, I think I was renaming stuff and didn't realize I did this

natjoe4 (approved):

src/scripts/getConnectathonBundles.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Uploads all the resources from the connectathon measure bundles into the
+ * database.
+ *
+ * TODO: Currently configured for fhir401 connectathon measure bundles,
+ * but may want to expand to other measure bundle providers in the future.
+ */
+async function main() {
+  getBundleFiles(connectathonPath);
```
hossenlopp, line 39: Suggest adding a check to make sure the c-thon folder exists first. Giving a helpful error message to do the clone before running this script would be a nice addition.

src/scripts/getConnectathonBundles.js
```diff
@@ -0,0 +1,67 @@
...
+    // read each EXM bundle file
+    const data = fs.readFileSync(filePath, 'utf8');
+    if (data) {
+      const bundle = JSON.parse(data);
+      // retrieve each resource and insert into database
+      const uploads = bundle.entry.map(async res => {
+        try {
+          await createResource(res.resource, res.resource.resourceType);
+        } catch (e) {
+          console.log(e.message);
```
hossenlopp, line 54: It's very common that the shared Libraries or ValueSets will be in multiple measure bundles and will fail to insert (assuming indexes are configured) after the first copy is inserted. This is a normal thing that shouldn't spam the output. Either catch this `duplicate key error` and not bother printing it out unless it is a `Measure` resource or check for existence before attempting the insert.

natjoe4 (approved):

hossenlopp (changes requested):
Async stuff looks all correct and it works. Just have some concerns about the error states.

src/scripts/getConnectathonBundles.js
sarahmcdougall: Good point! That should definitely be added

natjoe4 (approved):

src/scripts/getConnectathonBundles.js
```diff
@@ -0,0 +1,76 @@
...
+    const data = fs.readFileSync(filePath, 'utf8');
+    if (data) {
+      const bundle = JSON.parse(data);
+      // retrieve each resource and insert into database
+      const uploads = bundle.entry.map(async res => {
+        try {
+          await createResource(res.resource, res.resource.resourceType);
+        } catch (e) {
+          // ignore duplicate key errors for Libraries, ValueSets
+          if (res.resource.resourceType === 'Measure') {
```
hossenlopp, line 61: This hides ALL errors unless they are on Measure. Suggest checking on error code. `11000` is the code for duplicate key.
```suggestion
          if (e.code !== 11000 || res.resource.resourceType === 'Measure') {
```

hossenlopp (changes requested):
Shouldn't filter out ALL error messages if they aren't on `Measure`.

src/scripts/getConnectathonBundles.js
sarahmcdougall: Thank you!!

hossenlopp (approved):
Looks good!

natjoe4 (approved):

PR #22, opened 2021-09-28
added param/error handling for evaluate-measure

sarahmcdougall:
# Summary
Additional parameter validation and handling was added to the `$evaluate-measure` operation so that more input parameters are supported and properly handled.

## New behavior
Previously, the `$evaluate-measure` operation simply checked for the presence of the subject parameter and did not support all possible report types. Now, errors are thrown when required parameters are not included in the http request, when unsupported parameters (`measure`, `practitioner`, `lastReceivedOn`) are included in the request, when an unsupported `reportType` is provided (`population` or `subject-list`), and when an invalid `reportType` is provided.

## Code changes
The `validateEvalMeasureParams` function was created in `measureOperationsUtils.js` to throw `ServerError`s is faulty input is given. A `ServerError` may be thrown with all the missing required params, all the included but unsupported params, an included but unsupported `reportType`, and an included but invalid `reportType`. This function is called in `evaluateMeasure` before calling `calculateMeasureReports`. 

# Testing guidance
* POST the EXM130 bundle to `http://localhost:3000/4_0_0/`. You should then have a Measure in the database with id `measure-EXM130-7.3.000`, as well as Patients with IDs `numer-EXM130` and `denom-EXM130`. 
* Run `GET http://localhost:3000/4_0_0/Measure/measure-EXM130-7.3.000/$evaluate-measure?subject=numer-EXM130&periodStart=2019-01-01&periodEnd=2019-12-31` and check that you get the desired MeasureReport.

Then, check that ServerErrors are thrown when:
* You are missing one or more required params (`periodStart`, `periodEnd`, `subject`)
* You have included one or more unsupported params (`measure`, `practitioner`, `lastReceivedOn`)
* You have included an unsupported reportType (for example, `subject-list`, which has not been implemented yet in `fqm-execution`)
* You have included an invalid reportType (anything aside from `individual`, `subject-list`, `population`)

**NOTE**: As someone who hasn’t done much work in `fqm-execution` with the calculator functions for MeasureReports, please check that I did not miss any important parameter handling (I tried to base my work off of this: https://www.hl7.org/fhir/operation-measure-evaluate-measure.html). If other situations fall under the scope of this task, please let me know and I will add them.

Also, I am a little confused about when we are supposed to use `subject`, `subject-list` and `population` versus `individual`, `subject-list`, or `summary`, so if I used the wrong report types for `$evaluate-measure` please let me know.

### DIFF
--- a/src/util/bundleUtils.js
+++ b/src/util/bundleUtils.js
@@ -266,9 +266,73 @@ function replaceReferences(entries) {
   return newEntries;
 }
 
+/**
+ * Checks that the parameters input to $evaluate-measure are valid. Throws a ServerError
+ * for missing parameters, the use of unsupported parameters, and the use of unsuppported
+ * report types.
+ * @param {Object} req http request object
+ */
+function validateEvalMeasureParams(req) {
+  const REQUIRED_PARAMS = ['periodStart', 'periodEnd', 'subject'];
+  const UNSUPPORTED_PARAMS = ['measure', 'practitioner', 'lastReceivedOn'];
+
+  const missingParams = REQUIRED_PARAMS.filter(key => !req.query[key]);
+  const includedUnsupportedParams = UNSUPPORTED_PARAMS.filter(key => req.query[key]);
+
+  // returns all required params that are missing in the http request
+  if (missingParams.length > 0) {
+    throw new ServerError(null, {
+      statusCode: 400,
+      issue: [
+        {
+          severity: 'error',
+          code: 'BadRequest',
+          details: {
+            text: `Missing required parameters: ${missingParams.join(', ')} for $evaluate-measure`
+          }
+        }
+      ]
+    });
+  }
+  // returns all unsupported params that are included in the http request
+  if (includedUnsupportedParams.length > 0) {
+    throw new ServerError(null, {
+      statusCode: 400,
+      issue: [
+        {
+          severity: 'error',
+          code: 'BadRequest',
+          details: {
+            text: `The following parameters were included and are not supported for $evaluate-measure: ${includedUnsupportedParams.join(
+              ', '
+            )}`
+          }
+        }
+      ]
+    });
+  }
+
+  // returns unsupported report type that is included in the http request
+  if (!['individual', 'summary', undefined].includes(req.query.reportType)) {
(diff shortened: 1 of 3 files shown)


src/util/bundleUtils.js
```diff
@@ -266,9 +266,73 @@ function replaceReferences(entries) {
   return newEntries;
 }
 
+/**
+ * Checks that the parameters input to $evaluate-measure are valid. Throws a ServerError
+ * for missing parameters, the use of unsupported parameters, and the use of unsuppported
+ * report types.
+ * @param {Object} req http request object
+ */
+function validateEvalMeasureParams(req) {
```
mgramigna, line 275: I think this bundleUtils.js file is getting a little crowded now. And this function has nothing to do with a bundle. Can we move this to its own file? Maybe measureOperationsUtils or something?

src/util/bundleUtils.js
```diff
@@ -266,9 +266,73 @@ function replaceReferences(entries) {
...
+              ', '
+            )}`
+          }
+        }
+      ]
+    });
+  }
+
+  // returns unsupported report type that is included in the http request
+  if (!['individual', 'summary', undefined].includes(req.query.reportType)) {
```
mgramigna, line 316: Right now we only support individual measure reports (or omitted reportType param entirely).

Can we have separate logic for saying that ~summary~ population and subject-list measure reports are not yet supported by the server? Maybe we can use the not implemented error rather than the bad request. It makes sense to keep the bad request error in the case where the reportType provided is not one of the valid ones detailed by the spec.

mgramigna (changes requested):
Looks good. Made 2 inline comments, and looks like the [spec](https://www.hl7.org/fhir/operation-measure-evaluate-measure.html) says that the proper codes for reportType are `population`, `individual`, `subject-list` and nothing, so I guess we should be using `population` instead of `summary`, and also add `subject-list` where relevant

src/util/bundleUtils.js
sarahmcdougall: good point, I agree that it should be in a new file

mgramigna (approved):

PR #25, opened 2021-10-04
Bulk submit

natjoe4:
# Summary

## Code changes

# Testing guidance


test/bulkstatus.service.test.js
```diff
@@ -0,0 +1,59 @@
...
+        expect(response.headers['retry-after']).toEqual('120');
+      });
+  });
+  test('check 200 returned for completed request', async () => {
+    await supertest(server.app)
+      .get('/4_0_0/bulkstatus/COMPLETED_REQUEST')
+      .expect(200)
+      .then(response => {
+        expect(response.headers.expires).toBeDefined();
+        expect(response.headers['content-type']).toEqual('application/json; charset=utf-8');
```
natjoe4, line 25: should just be application/json here I think, but not sure if this matters

test/bulkstatus.service.test.js
```diff
@@ -0,0 +1,59 @@
...
+        expect(response.body).toBeDefined();
+      });
+  });
+  test('check 500 and error returned for failed request with known error', async () => {
+    await supertest(server.app)
+      .get('/4_0_0/bulkstatus/KNOWN_ERROR_REQUEST')
+      .expect(500)
+      .then(response => {
+        expect(response.body.issue[0].code).toEqual('ErrorCode');
+        expect(response.body.issue[0].details.text).toEqual('Known Error Occurred!');
```
sarahmcdougall, line 35: Based on [this documentation](https://github.com/smart-on-fhir/bulk-import/blob/master/import-pnp.md#bulk-data-import-status-request-data-provider-to-data-consumer) , do we need to check that the content-type is application/json+fhir when we get a 400 or 500 status code?

sarahmcdougall (changes requested):
Good work! The code seems intuitive to me and the tests look good and cover all the edge cases. Not a lot of changes to be made. I think we can get rid of this line: https://github.com/projecttacoma/deqm-test-server/blob/fd72a0d601620779c579e303e73df0bcb3284abf/src/services/measure.service.js#L153 because we are using all the vars now. Other than that, I just have a q below about checking for content-type for the error tests.

test/bulkstatus.service.test.js
natjoe4: might as well. Good call

PR #29, opened 2021-10-05
update package lock and ci

mgramigna:
for some reason, package-lock.json wasn't in sync with package.json and this caused npm ci to fail and therefore all of gh actions to fail

Also fixed the indenting in ci.yml

To test this, validate the yml syntax/content in `ci.yml`, and ensure that running `npm install` generates no package-lock diffs, and that the command `npm ci` works

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ jobs:
       - run: npm run check
 
   coverage:
-     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


sarahmcdougall (approved):
The indentation is correct now, and I had no issues with running `npm install` and `npm ci`. `npm install` doesn't create any package-lock diffs, so everything should be good now. Thanks for the fix!

PR #35, opened 2021-10-15
Allow "population" arg to be passed in to $evaluate-measure requests

natjoe4:
# Summary
Allows users to pass in "population" as a report type to run summary reports for `$evaluate-measure` requests. 

## New behavior
The "subject" parameter is no longer required for `$evaluate-measure` requests, as long as the `reportType` field in the query is set to `"population"`. Setting `reportType` to `population` will pass `reportType: summary` in to the fqm-execution calculateMeasureReport function

## Code changes
- Updated the validateEvalMeasureParams function in src/util/measureOperationsUtils.js to remove `subject` from required parameters if `reportType` is set to `"population"`
- Added code to pull all patients and their respective bundles from the DB to pass in to fqm-execution's calculateMeasureReport function if `reportType` is set to `"population"`
- Added testing to ensure correct response codes and function calls are made through this new workflow

# Testing guidance
- run `npm run test` to ensure all tests pass 
- upload the EXM130 bundle to the test server by sending a `POST` request to `http://localhost:3000/4_0_0` with the bundle in the request body
- send a `GET` request to `http://localhost:3000/4_0_0/Measure/measure-EXM130-7.3.000/$evaluate-measure?periodStart=01-01-2019&periodEnd=01-01-2020&reportType=population` and ensure that you receive a 200 response code and an output bundle of type "summary"
- continue this process with other Connectathon bundles, making sure to update the measureId in the url before sending the `GET` request

### DIFF
--- a/src/util/measureOperationsUtils.js
+++ b/src/util/measureOperationsUtils.js
@@ -76,7 +76,7 @@ function validateEvalMeasureParams(req) {
     });
   }
 
-  if (!req.query.subject && !req.query.reportType === 'population') {
+  if (!req.query.subject && req.query.reportType !== 'population') {
     throw new ServerError(null, {
       statusCode: 400,
       issue: [


src/util/measureOperationsUtils.js
```diff
@@ -76,6 +75,21 @@ function validateEvalMeasureParams(req) {
       ]
     });
   }
+
+  if (!req.query.subject && !req.query.reportType === 'population') {
```
sarahmcdougall, line 79:
```suggestion
  if (!req.query.subject && req.query.reportType !== 'population') {
```

sarahmcdougall: It wasn't throwing the error with `!req.query.reportType`

PR #36, opened 2021-10-15
Dynamic Export URL Handling

sarahmcdougall:
# Summary
Made updates so that `bulk-data-utilities` dynamically takes an export URL to a FHIR reference server in order to map data requirements into an export request.

## New Behavior
When a user specifies a single, valid export URL in the initial request body, the bulk import operation uses it in its call to `bulk-data-utilities` for mapping the data requirements into an export request. In other cases (no URL is supplied, the URL does not refer to a valid FHIR reference server, multiple export URLs are supplied), a `ServerError` is thrown.

## Code Changes
Within `measureOperationsUtils`, code was added to retrieve the export URL from the request body parameters and throw a `ServerError` in the case that a single, valid export URL `valueString` cannot be retrieved from the request body parameters. This code is called within `bulkImport`. Once the `exportURL` is retrieved, the call is made to `bulk-data-utilities` to map the data requirements into an export request. If the request cannot be made, then `bulk-data-utilities` sends back an `AxiosError` message that gets wrapped into a `ServerError`. This error would be thrown in `queryBylkDataServer()` within `bulk-data-utilities`.

After the `bulk-data-utilities` PR gets approved, we will need to change the dependency in `package.json` to point to the main branch.

# Testing Guidance
There are two ways to test this PR. First, ensure that all the unit tests pass. The unit tests cover the various error cases. Then, try sending a request in the test server. POST a measure bundle to `http://localhost:3000/4_0_0/`. Then, create a FHIR Parameters object that includes the `exportURL` in the `parameter` array. For example,
```
{
     "name": "exportURL",
     "valueString": "https://bulk-data.smarthealthit.org/eyJlcnIiOiIiLCJwYWdlIjoxMDAwMCwiZHVyIjoxMCwidGx0IjoxNSwibSI6MSwic3R1IjozLCJkZWwiOjB9/fhir"
}
```

POST this FHIR Parameters object to `http://localhost:3000/4_0_0/Measure/<measure-id>/$submit-data` or `http://localhost:3000/4_0_0/Measure/$submit-data` and make sure the “prefer:” “respond-async” header is present so that $submit-data calls the `bulkImport` function. Check that the ndjson url data is returned with the `202 Accepted` status. Then, try
* Omitting the entire object
* Including `”name”: “exportURL”` but not the `valueString`
* Including more than one object containing an `exportURL`
* Including an invalid `valueString` (for example, `not-real` instead of the SMART on FHIR reference server URL)

You should receive a `ServerError` for each of these cases.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3455,8 +3455,8 @@
       "dev": true
     },
     "bulk-data-utilities": {
-      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#7979962c5d6ff798ea8a3fc3fa2e561554d2f282",
-      "from": "git+https://github.com/projecttacoma/bulk-data-utilities.git#export-url",
+      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#32be10a86f6fc7386b26ae93df4a16d2944936d4",
+      "from": "git+https://github.com/projecttacoma/bulk-data-utilities.git",
       "requires": {
         "axios": "^0.21.1",
         "fqm-execution": "^0.8.0",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@asymmetrik/fhir-qb-mongo": "^0.12.3",
     "@asymmetrik/node-fhir-server-core": "^2.2.3",
     "axios": "^0.21.4",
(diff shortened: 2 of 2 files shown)


mgramigna:
Code is good here. Just need to update the BDU dependency to be the master branch now that https://github.com/projecttacoma/bulk-data-utilities/pull/9 is merged

PR #43, opened 2021-11-01
Unit tests

sarahmcdougall:
# Summary
The threshold in GitHub Actions was increased to 60. Additional unit tests and accompanying fixtures were added to the test server.

## New behavior
Behavior should remain unchanged.

## Code changes
Unit tests were added for the following files: `base.service.js`, `bundle.service.test.js`, `bundleUtiles.js`, `measure.service.js`, `measureOperationsUtils.js`, `transactionBundle.js`.
An error in `measure.service.js` is never reached (other ServerErrors are thrown in `getMeasureBundleFromId()`, so the error was removed.

# Testing guidance
Run `npm run test` and ensure that all the unit tests pass. Please check the unit tests to make sure their logic is correct.

Run `npm run test:coverage` to view the test coverage for all statements, branches, functions, and lines. Check that the test coverage percentages look good to you.

Please let me know if you think any of the uncovered lines should be covered, or if a file's test organization should be restructured using describe blocks.

Note that in `measure.service.js`, the `executePingandPull` code has not been unit tested because of the limitations mentioned in https://github.com/projecttacoma/deqm-test-server/pull/34.

### DIFF
--- a/test/measureOperationsUtils.test.js
+++ b/test/measureOperationsUtils.test.js
@@ -1,4 +1,8 @@
-const { checkRequiredParams } = require('../src/util/measureOperationsUtils');
+const {
+  checkRequiredParams,
+  validateEvalMeasureParams,
+  validateCareGapsParams
+} = require('../src/util/measureOperationsUtils');
 
 describe('measureOperationsUtils functions', () => {
   test('check checkRequiredParams throws error on missing params', () => {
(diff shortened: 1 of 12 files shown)


test/measureOperationsUtils.test.js
```diff
@@ -18,4 +22,113 @@ describe('measureOperationsUtils functions', () => {
     const REQUIRED_PARAMS = ['test', 'test2'];
     checkRequiredParams(req, REQUIRED_PARAMS, 'test');
   });
+
```
natjoe4, line 25: I would add a test here as well that validateEvalMeasureParams doesn't throw an error when passed valid parameters

test/measureOperationsUtils.test.js
```diff
@@ -18,4 +22,113 @@ describe('measureOperationsUtils functions', () => {
...
+    try {
+      validateEvalMeasureParams(MISSING_SUBJECT_REQ);
+    } catch (e) {
+      expect(e.statusCode).toEqual(400);
+      expect(e.issue[0].details.text).toEqual(
+        `Must specify subject for all $evaluate-measure requests with reportType: individual`
+      );
+    }
+  });
+
```
natjoe4, line 77: I would add a test here to check that validateCareGapsParams doesn't throw an error when passed valid parameters

test/measureOperationsUtils.test.js
natjoe4, line 25: I would add a test here to check that validateEvalMeasureParams doesn't throw an error when passed valid parameters

test/measureOperationsUtils.test.js
natjoe4, line 77: I would add a test here to check that validateCareGapsParams doesn't throw an error when passed valid parameters

natjoe4 (approved):
Awesome work. Looks good to me!

PR #44, opened 2021-11-02
updated to not break on build

natjoe4:
# Summary
Quick patch that allows the dockerfile to build the image without crashing in the absence of a docker_ssl_setup.sh file
## New behavior
The docker image will not crash on build in the absence of a docker_ssl_setup.sh file
## Code changes
Added a wildcard to the end of docker_ssl_setup.sh on the copy step and now copy the readme.md before it to ensure the build doesn't crash
# Testing guidance
run `docker build -t deqm-test-server .` and ensure it builds, then delete `docker_ssl_setup.sh` and run the same command again and ensure that it fails on the npm install with `exit 1`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,8 @@ WORKDIR /usr/src/app
 
 
 # Run a custom ssl_setup script if available
-COPY ./docker_ssl_setup.sh ./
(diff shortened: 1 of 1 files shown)


mgramigna (approved):
Confirmed that this workaround causes a successful build without the ssl setup script being present

PR #46, opened 2021-11-03
Inclusion of X-Provenance header

sarahmcdougall:
# Summary
When a POST/PUT request is sent (assuming it contains the `Provenance` header), the `Provenance` `target` attribute is populated with the reference(s) to the ID(s) that the server is using for the resource(s) created via the POST/PUT request.

## New behavior
Now, POST/PUT requests are expected to contain a `Provenance` header, which is a JSON object that is clinically significant for authenticity and reliability. Right now, we are concerned that the Provenance header is present with the `Provenance` `resourceType`, that its target attribute is not initially specified, and that our POST/PUT requests create and populate the target attribute with references to the IDs that the server is using for a created resource.

The `target` attribute is populated for `baseCreate`, `baseUpdate`, and `uploadTransactionBundle` because they send POST/PUT requests. 

## Code changes
Functions were created in `base.service.js` to check that the included provenance header is valid and to populate the target attribute of the provenance header appropriately. The `baseUpdate`, `baseCreate`, and transaction bundle functions were updated to check for the provenance header and to populate it with the necessary references. In the case of the transaction bundle upload, all the creates/updates leave us with a bunch of `target` arrays, which we must combine together when populating the transaction bundle `target` with all the reference objects.

Some of the unit tests were updated so that the `x-provenance` header is present - otherwise those tests would now fail.

# Testing guidance
Run the unit tests and ensure that they pass. Try doing a standard transaction bundle upload with the EXM130 bundle and add the header `X-Provenance: { "resourceType": "Provenance”}`. After sending the request, check the response header in the Insomnia UI. It should have an X-Provenance header that is populated with the `resourceType` and a target array full of the references.

### DIFF
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -83,11 +83,12 @@ const qb = new QueryBuilder({
 const baseCreate = async ({ req }, resourceType) => {
   logger.info(`${resourceType} >>> create`);
   checkContentTypeHeader(req.headers);
-  // console.log(req.headers);
-  // checkProvenanceHeader(req.headers);
+  checkProvenanceHeader(req.headers);
+  const res = req.res;
   const data = req.body;
   //Create a new id regardless of whether one is passed
   data['id'] = uuidv4();
+  populateProvenanceTarget(req.headers, res, [{ reference: `${resourceType}/${data.id}` }]);
 
   return createResource(data, resourceType);
 };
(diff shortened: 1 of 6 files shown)


natjoe4 (approved):
No Notes. This looks great! Nice work

PR #48, opened 2021-11-08
made x-provenance optional

sarahmcdougall:
# Summary
Updated code so that the X-Provenance header is optional.

## New behavior
It makes more sense right now for the X-Provenance header to be optional for POST/PUT requests until we test X-Provenance more. 

See [the original PR](https://github.com/projecttacoma/deqm-test-server/pull/46) to learn more about X-Provenance.

## Code changes
Changes were made to the base create, base update, and transaction bundle upload to make the X-Provenance header optional. The validation handling and unit tests were updated to reflect these changes.

# Testing guidance
Ensure that the unit tests run. Run the same tests outlined in [the original PR](https://github.com/projecttacoma/deqm-test-server/pull/46) to make sure that functionality is good both with and without the header.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2680,10 +2680,10 @@
       "dev": true
     },
     "bulk-data-utilities": {
-      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#fb04a13d49357b74e6e8fed55e97d44a4b490c2d",
+      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#32be10a86f6fc7386b26ae93df4a16d2944936d4",
       "from": "git+https://github.com/projecttacoma/bulk-data-utilities.git",
       "requires": {
-        "axios": "^0.21.2",
+        "axios": "^0.21.1",
         "fqm-execution": "^0.8.0",
         "path": "^0.12.7",
         "sqlite": "^4.0.23",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2187,9 +2187,9 @@
       "integrity": "sha1-ml9pkFGx5wczKPKgCJaLZOopVdI="
     },
     "asn1": {
-      "version": "0.2.4",
-      "resolved": "https://registry.npmjs.org/asn1/-/asn1-0.2.4.tgz",
-      "integrity": "sha512-jxwzQpLQjSmWXgwaCZE9Nz+glAG01yF1QnWgbhGwHI5A6FRIEY6IVqtHhIepHqI7/kyEyQEagBC5mBEFlIYvdg==",
+      "version": "0.2.6",
+      "resolved": "https://registry.npmjs.org/asn1/-/asn1-0.2.6.tgz",
+      "integrity": "sha512-ix/FxPn0MDjeyJ7i/yoHGFt/EX6LyNbxSEhPPXODPL+KB0VPk86UYfL0lMdy+KCnv+fmvIzySwaK5COwqVbWTQ==",
       "optional": true,
       "requires": {
         "safer-buffer": "~2.1.0"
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -2680,10 +2680,10 @@
       "dev": true
     },
     "bulk-data-utilities": {
-      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#fb04a13d49357b74e6e8fed55e97d44a4b490c2d",
+      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#32be10a86f6fc7386b26ae93df4a16d2944936d4",
```
mgramigna, line 2683: it looks like the commit hash of the bulk-data-utilities branch has been downgraded by a few commits, can you undo this and re-npm install just to make sure that it doesn't actually generate any diffs? Assuming this was accidental

test/bundle.service.test.js
```diff
@@ -46,10 +46,11 @@ describe('Test transaction bundle upload', () => {
       .send(testBundle)
       .set('Accept', 'application/json+fhir')
       .set('content-type', 'application/json+fhir')
-      .set('x-provenance', '{ "resourceType": "Provenance"}')
+      .set('x-provenance', '{"resourceType": "Provenance"}')
       .expect(200)
       .then(async response => {
         // Check the response
+        console.log(response.headers['x-provenance']);
```
mgramigna, line 53:
```suggestion
```

test/bundle.service.test.js
```diff
@@ -35,7 +35,7 @@ describe('uploadTransactionBundle Server errors', () => {
     }
   });
 });
-describe('Test transaction bundle upload', () => {
+describe.only('Test transaction bundle upload', () => {
```
mgramigna, line 38:
```suggestion
describe('Test transaction bundle upload', () => {
```

src/services/base.service.js
```diff
@@ -83,13 +83,14 @@ const qb = new QueryBuilder({
...
   logger.info(`${resourceType} >>> create`);
   checkContentTypeHeader(req.headers);
-  checkProvenanceHeader(req.headers);
-  const res = req.res;
   const data = req.body;
   //Create a new id regardless of whether one is passed
   data['id'] = uuidv4();
-  populateProvenanceTarget(req.headers, res, [{ reference: `${resourceType}/${data.id}` }]);
-
+  if (Object.keys(req.headers).includes('x-provenance')) {
```
mgramigna, line 89: Nitpick: we can check the key with an `O(1)` lookup rather than having to do a search in an array with `includes` (`O(n)`)

```suggestion
  if (req.headers['x-provenance']) {
```

src/services/base.service.js
```diff
@@ -233,7 +233,11 @@ const baseUpdate = async (args, { req }, resourceType) => {
       ]
     });
   }
-  populateProvenanceTarget(req.headers, res, [{ reference: `${resourceType}/${args.id}` }]);
+  if (Object.keys(req.headers).includes('x-provenance')) {
```
mgramigna, line 236: Same

```suggestion
  if (req.headers['x-provenance']) {
```

src/services/bundle.service.js
```diff
@@ -79,24 +83,37 @@ async function uploadTransactionBundle(req, res) {
       ]
     });
   }
-  checkProvenanceHeader(req.headers);
+  let xprovenanceIncluded;
+  if (Object.keys(req.headers).includes('x-provenance')) {
```
mgramigna, line 87: Same same 

```suggestion
  if (req.headers['x-provenance']) {
```

src/services/bundle.service.js
```diff
@@ -79,24 +83,37 @@ async function uploadTransactionBundle(req, res) {
...
-    });
+    if (xprovenanceIncluded) {
+      return axios[method.toLowerCase()](destinationUrl, entry.resource, {
+        headers: { 'Content-Type': 'application/json+fhir', 'X-Provenance': req.headers['x-provenance'] }
+      });
+    } else {
+      return axios[method.toLowerCase()](destinationUrl, entry.resource, {
+        headers: { 'Content-Type': 'application/json+fhir' }
+      });
+    }
```
mgramigna, line 104: Nitpick: Can we avoid the code duplication here by creating a `headers` variable and then adding the `X-Provenance` key/value only if `xprovenanceIncluded` is true? That way, we can only have one axios call, and just pass in that variable as the headers argument

mgramigna (changes requested):
Functionality looks good. A few code suggestions/nitpicks

package-lock.json
sarahmcdougall: good catch, yes this was accidental, not sure what happened

mgramigna (approved):

PR #49, opened 2021-11-12
Support for Patient/$everything

sarahmcdougall:
# Summary
This PR adds support for the `Patient/$everything` operation in the test server.

## New behavior
This new operation will return all the information related to one of more patients, with a response that is a `searchset` bundle. The operation can be used by sending a `GET` request to `[base]/Patient/$everything` or `[base]/Patient/[id]/$everything`. 

## Code changes
`patient.service.js` was created to cover basic CRUD operations and the `$everything` operation. Parameter validation was added to throw errors if any non-implemented parameters are provided in the request. Note that there are no required parameters for this operation.

`config.js` was updated to support the `$everything` route.

In `bundleutils.js`, the `getPatientDataBundle()` function was changed since we want to use it for the `$everything` operation as well as with measure operations like `$evaluate-measure`. However, the measure operations map to a collection bundle, whereas `$everything` maps to a searchset bundle. Two wrapper functions were created to account for the different bundle type scenarios.

Unit testing was added.
 

# Testing guidance
Ensure that the unit tests pass and provide adequate coverage of the added functionality. Then, POST a bundle containing a patient resource and resources related to the patient. Try sending a GET request and check that the correct response is returned. For example, I POSTed the EXM 130 bundle and then sent requests to
* `http://localhost:3000/4_0_1/Patient/$everything`
* `http://localhost:3000/4_0_1/Patient/numer-EXM130/$everything`
* `http://localhost:3000/4_0_1/Patient/$everything?start=2019-01-01` (will throw ServerError)

### DIFF
--- a/src/services/patient.service.js
+++ b/src/services/patient.service.js
@@ -1,6 +1,6 @@
-const { loggers } = require('@projecttacoma/node-fhir-server-core');
+const { ServerError, loggers } = require('@projecttacoma/node-fhir-server-core');
 const { baseCreate, baseSearchById, baseRemove, baseUpdate, baseSearch } = require('./base.service');
-const { getPatientDataBundle } = require('../util/bundleUtils');
+const { getPatientDataSearchSetBundle } = require('../util/bundleUtils');
 const { findResourcesWithQuery } = require('../util/mongo.controller');
 const logger = loggers.get('default');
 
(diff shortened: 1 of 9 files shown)


src/util/bundleUtils.js
```diff
@@ -188,15 +189,32 @@ async function getAllDependentLibraries(lib) {
...
 }
 
+async function getPatientDataCollectionBundle(patientId, dataRequirements) {
+  const data = await getPatientData(patientId, dataRequirements);
+  return mapResourcesToCollectionBundle(_.flattenDeep(data));
+}
+
+async function getPatientDataSearchSetBundle(patientId, args, req) {
+  const data = await getPatientData(patientId);
+  console.log(_.flattenDeep(data));
```
natjoe4, line 199: Do we want this console.log?

src/services/patient.service.js
```diff
@@ -0,0 +1,118 @@
...
+    const patients = await findResourcesWithQuery({}, 'Patient');
+    let patientBundles = patients.map(async p => {
+      return getPatientDataSearchSetBundle(p.id, args, req);
+    });
+
+    patientBundles = await Promise.all(patientBundles);
+    return patientBundles;
+  }
+};
+
```
natjoe4, line 85: This function could use a jsdoc

src/util/bundleUtils.js
```diff
@@ -188,15 +189,32 @@ async function getAllDependentLibraries(lib) {
   return results;
 }
```
natjoe4, line 191: The two below could also use a jsdoc

natjoe4 (changes requested):
This code all works for me as described. Nice Job! I left a couple small documentation comments :)

natjoe4 (approved):
This all looks great to me! Nice work :)

natjoe4 (approved):
Actually approved this time...

PR #50, opened 2021-11-16
Updated config and patient data logic for Patient/$everything

sarahmcdougall:
# Summary
The config was updated to support POST requests to Patient/$everything so that we can test the operation in the deqm test kit. Updates were made so that Patient/$everything returns a single bundle and not an array of bundles for each patient.

## New behavior
^
## Code changes
yes
# Testing guidance
Ensure that unit tests still pass and that calling `Patient/$everything` will give you a bundle, not an array of bundles.


mgramigna (approved):
sure

PR #53, opened 2021-12-01
Audit references

lmd59:
# Summary
Changes the creation of AuditEvents a bit from Nat's code since these were developed at the same time a little differently, but otherwise uses Nat's X-provenance mapping to create AuditEvents and adds references to the uploaded resources. Adds some tests to exercise resources in the request body and check for AuditEvent added.

## New behavior
References to the submitted resources (with appropriate resource location) are added to the audit event.

## Code changes
Creates a new pushToResource function for adding to arrays on a mongo resource.
Uses new `handleSubmitDataBundles` function to do appropriate bundle handling for submit data including creating the audit event.
Makes changes in places where `uploadTransactionBundle` to use `handleSubmitDataBundles` more generically instead.

# Testing guidance
Make sure mongo is running. Run tests with `npm run test`. Submit test requests with resources in the body and X-Provenance headers to http://localhost:3000/4_0_0/Measure/$submit-data
An example of parameters with resources can be found here: https://github.com/projecttacoma/deqm-test-server/blob/audit_references/test/fixtures/parametersObjs/paramNoExportResource.json
An example of xprovenance for the header can be found here: https://github.com/projecttacoma/deqm-test-server/blob/cd73371ea381b71edae05f4fd4c39105d5622c4d/test/fixtures/testProvenanceUtils.js#L22
Feel free to play around with the data to break.
Use robo 3t or a get request to http://localhost:3000/4_0_0/AuditEvent to look at the created AuditEvents and ensure that they have the correct references to the submitted resources
This should also work for a bulk submit data.

### DIFF
--- a/src/services/bundle.service.js
+++ b/src/services/bundle.service.js
@@ -50,12 +50,12 @@ const makeTransactionResponseBundle = (results, res, baseVersion, type, xprovena
 /**
  * Handles transaction bundles used for submit data.
  * Creates an audit event and uploads the transaction bundles.
- * @param {*} transactionBundles - an array of transactionBundles to handle
- * @param {*} req - an object containing the request body
+ * @param {Array} transactionBundles - an array of transactionBundles to handle
+ * @param {Object} req - an object containing the request body
  * @returns array of transaction-response bundle
  */
 async function handleSubmitDataBundles(transactionBundles, req) {
-  var auditID;
+  let auditID;
   const { base_version: baseVersion } = req.params;
   if (req.headers['x-provenance']) {
     checkProvenanceHeader(req.headers);
--- a/src/services/bundle.service.js
+++ b/src/services/bundle.service.js
@@ -4,6 +4,8 @@ const { ServerError, loggers, resolveSchema } = require('@projecttacoma/node-fhi
 const { v4: uuidv4 } = require('uuid');
 const { replaceReferences } = require('../util/bundleUtils');
 const { checkProvenanceHeader, populateProvenanceTarget } = require('../util/provenanceUtils');
+const { createResource, pushToResource } = require('../util/mongo.controller')
+const { createAuditEventFromProvenance } = require('../util/provenanceUtils');
 
 const logger = loggers.get('default');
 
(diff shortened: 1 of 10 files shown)


src/util/mongo.controller.js
```diff
@@ -62,6 +62,29 @@ const updateResource = async (id, data, resourceType) => {
...
+ * searches for a document and updates it by pushing to existing if found, creates it if not
+ * @param {*} id id of resource to be updated
+ * @param {*} data the new data to push in the document
+ * @param {*} resourceType the collection the document is in
+ * @returns the id of the updated/created document
+ */
+const pushToResource = async (id, data, resourceType) => {
+  const collection = db.collection(resourceType);
+
+  const results = await collection.findOneAndUpdate({ id: id }, { $push: data });
```
lmd59, line 75: pushing multiple requires an $each i.e. data = { scores: { $each: [ 90, 92, 85 ] } }
should $each be passed in as data, or is that bad modularization? (currently being passed in as data)

src/services/bundle.service.js
```diff
@@ -45,6 +47,41 @@ const makeTransactionResponseBundle = (results, res, baseVersion, type, xprovena
...
 
+/**
+ * Handles transaction bundles used for submit data.
+ * Creates an audit event and uploads the transaction bundles.
+ * @param {*} transactionBundles - an array of transactionBundles to handle
+ * @param {*} req - an object containing the request body
+ * @returns array of transaction-response bundle
+ */
+async function handleSubmitDataBundles(transactionBundles, req) {
+  var auditID;
```
mgramigna, line 58:
```suggestion
  let auditID;
```

src/services/measure.service.js
```diff
@@ -217,7 +211,7 @@ const bulkImport = async (args, { req }) => {
 /**
  * Calls the bulk-data-utilities wrapper function to get data requirements for the passed in measure, convert those to
  * export requests from a bulk export server, then retrieve ndjson from that server and parse it into valid transaction bundles.
- * Finally, uploads the resulting transaction bundles to the server and updateed the bulkstatus endpoint
+ * Finally, uploads the resulting transaction bundles to the server and updates the bulkstatus endpoint
```
mgramigna, line 214: lol nice

src/util/bundleUtils.js
```diff
@@ -269,7 +269,7 @@ async function getPatientData(patientId, dataRequirements) {
  * Modify the request type to PUT after forcing the IDs. This will not affect return results, just internal representation
  *
  * @param {Array} entries array of bundle entries
- * @returns new array of entries with replaced reverences
+ * @returns new array of entries with replaced references
```
mgramigna, line 272: lol nice

src/util/mongo.controller.js
```diff
@@ -62,6 +62,29 @@ const updateResource = async (id, data, resourceType) => {
   return { id: results.value.id, created: false };
 };
 
+/**
+ * searches for a document and updates it by pushing to existing if found, creates it if not
+ * @param {*} id id of resource to be updated
+ * @param {*} data the new data to push in the document
+ * @param {*} resourceType the collection the document is in
```
mgramigna, line 69:
```suggestion
 * @param {string} id id of resource to be updated
 * @param {Object} data the new data to push in the document
 * @param {string} resourceType the collection the document is in
```

src/util/mongo.controller.js
mgramigna: I think the way you've structured this is fine. We have other functions in the mongo controller that accept arbitrary mongo queries as arguments and it's been pretty flexible and beneficial, so I'm fine with leaving the arguments/function call how you've written it here

src/util/mongo.controller.js
```diff
@@ -62,6 +62,29 @@ const updateResource = async (id, data, resourceType) => {
...
+  // If the document cannot be created with the passed id, Mongo will throw an error
+  // before here, so should be ok to just return the passed id
+  if (results.value === null) {
+    // null value indicates a newly created document
+    return { id: id, created: true };
+  }
+
+  // value being present indicates an update, so set created flag to false
+  return { id: results.value.id, created: false };
+};
```
mgramigna, line 86: Do we need this block of code? Since we're not doing any upserting, I don't think we'll ever reach this `if` since it will only be updating existing documents.

We also don't make use of the return value at all, so I think it's okay for this function to be void and just do the `findOneAndUpdate` with no return value

src/services/bundle.service.js
```diff
@@ -45,6 +47,41 @@ const makeTransactionResponseBundle = (results, res, baseVersion, type, xprovena
   return { bundle, bundleProvenanceTarget: bundleProvenanceTarget.flat() };
 };
 
+/**
+ * Handles transaction bundles used for submit data.
+ * Creates an audit event and uploads the transaction bundles.
+ * @param {*} transactionBundles - an array of transactionBundles to handle
+ * @param {*} req - an object containing the request body
```
mgramigna, line 54:
```suggestion
 * @param {Array} transactionBundles - an array of transactionBundles to handle
 * @param {Object} req - an object containing the request body
```

mgramigna (changes requested):
Code looks good and works. I just have some really small comments

test/bundle.service.test.js
```diff
@@ -59,3 +60,41 @@ describe('Test transaction bundle upload', () => {
     await cleanUpDb();
   });
 });
+
+describe('Test handle submit data bundle', () => {
+  beforeAll(async () => {
+    await client.connect();
+  });
+
+  test('Submit data bundle with resources creates AuditEvent with resources', async () => {
```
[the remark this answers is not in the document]
sarahmcdougall: I think it's fine to have both of these tests even if there's a little bit of overlap. It's always better to have more tests, and the tests are focused on different parts of the functionality (successful submit-data request v. audit event being created).

sarahmcdougall (commented):
Great job integrating your AuditEvent work with Nat's! Code looks clean and functions as expected. Matt already caught all the potential changes that I saw.

src/util/mongo.controller.js
lmd59: Good call! I think my initial conception was that this could still work for a database document that doesn't exist yet, but that actually fails, so I'm going to make these changes and add that comment to the function docs.

mgramigna (approved):

sarahmcdougall (approved):

PR #54, opened 2021-12-01
Import endpoint

natjoe4:
# Summary
Exposed a `$import` endpoint for carrying out non-measure-specific bulk import

## New behavior
We can now send generic bulk import requests to the `$import` endpoint. These requests will export all the resources on  bulk export server referenced by the passed-in url

## Code changes
- Added src/services/import.service.js
- Added src/services/import.controller.js
- Exposed new `$import` endpoint
- Reconfigured existing `$import` code to work with updated bulk-data-utilities functions
- Moved existing `$import`-related functions to src/import.service.js
- Added unit tests

# Testing guidance
- Run all unit tests
- Testing this code is a little involved due to reference errors in both our bulk-export-server and the smarthealthIT server
- Open your local copy of bulk-export-server and change the `PORT` variable in .env to `8080`
- Boot up the bulk-export-server with `npm run start`
- Open your local copy of bulk-data-utilities and change the url in the `retrieveAllBulkData` function on line 149 of src/RequirementsQuery from ``const url = `${exportURL}/$export`;`` to ``const url = `${exportURL}/$export?_type=Patient`;``
- Back in DEQM-test-server, run `npm link {LOCAL_PATH_TO_BULK_DATA_UTILITIES_REPO}`
- Now, boot up the test server with `npm run start`
- Send a `POST` request to `http://localhost:3000/4_0_1/$import` with the following body `{
  "resourceType": "Parameters",
  "parameter": [
		{
     "name": "exportURL",
     "valueString": "http://localhost:8080"
}]}`
- You should receive a `202:Accepted` status with a `content-location` header
- Send a `GET` request to `localhost:3000{INSERT CONENT LOCATION HEADER HERE}`
- If you receive `202:Accepted`, repeat the request
- You should eventually expect to receive `200` code with an OperationOutcome returned with one url representing the ndjson file of patients exported from the server

### DIFF
--- a/test/import.service.test.js
+++ b/test/import.service.test.js
@@ -16,6 +16,12 @@ describe('Testing $import with no specified measure bundle', () => {
   beforeEach(async () => {
     await client.connect();
   });
+
+  /*
+   * Skipped purposely for now
+   * TODO: Once job queue is implemented, unskip this!
+   */
+
   test.skip('Returns 202 on Valid Request', async () => {
     await supertest(server.app)
       .post('/4_0_1/$import')
--- a/test/import.service.test.js
+++ b/test/import.service.test.js
@@ -1,0 +1,3 @@
+describe('Testing Bulk Import with no Specified Measure Bundle', () => {
+  test('Returns 200');
+});
--- a/test/import.service.test.js
+++ b/test/import.service.test.js
@@ -16,7 +16,7 @@ describe('Testing $import with no specified measure bundle', () => {
   beforeEach(async () => {
     await client.connect();
   });
-  test('Returns 202 on Valid Request', async () => {
+  test.skip('Returns 202 on Valid Request', async () => {
     await supertest(server.app)
       .post('/4_0_1/$import')
       .send(validParam)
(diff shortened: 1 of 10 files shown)


sarahmcdougall (changes requested):
I just have a few comments, but overall the functionality is good and the tests pass. After testing everything, I ended up with a bunch of untracked files in my root directory, and their names correspond to the client IDs for the import request. Not sure if this is a side effect of connecting to the bulk export server or if it's just something happening on my end.

Also, I think it would be worthwhile to update the README to reflect the non-measure specific bulk import.

test/import.service.test.js
```diff
@@ -0,0 +1,62 @@
...
+const { cleanUpDb } = require('./populateTestData');
+
+const config = buildConfig();
+const server = initialize(config);
+
+describe('Testing $import with no specified measure bundle', () => {
+  beforeEach(async () => {
+    await client.connect();
+  });
+  test.skip('Returns 202 on Valid Request', async () => {
```
sarahmcdougall, line 19:
```suggestion
  test('Returns 202 on Valid Request', async () => {
```

sarahmcdougall (changes requested):

test/import.service.test.js
natjoe4: This is actually skipped on purpose for now due to jest race conditions. I'll add a comment!

sarahmcdougall (approved):

PR #58, opened 2021-12-10
Type check

natjoe4:
# Summary
Added resource type checks for Create, Update, Transaction bundle upload and $submit-data

## New behavior
The test server will now throw 404 errors when the resource type included in the request body of a resource is not in our supported resources list

## Code changes
- Added a utility function to check the resourceType of a request body against our supported resources list
- Added code to run above function during Create, Update, Transaction bundle upload and $submit-data operations
- Updated Audit Event code to omit resources which failed to be Created/Updated during a transaction bundle upload or $submit data request
- Added testing for custom functions and updated existing testing to account for additions

# Testing guidance
- Try using the Create, Update, txn bundle upload, and $submit-data functionality using request bodies that contain invalid resource types and ensure that the error output is clear and appropriate
- Check that no other functionality is disturbed.

### DIFF
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -218,7 +218,9 @@ const baseUpdate = async (args, { req }, resourceType) => {
   logger.info(`${resourceType} >>> update`);
   checkContentTypeHeader(req.headers);
   const data = req.body;
-  checkSupportedResource(data.resourceType);
+  if (data.resourceType !== undefined) {
+    checkSupportedResource(data.resourceType);
+  }
   //The user passes in an id in the request body and it doesn't match the id arg in the url
   //or user doesn't pass in body
   if (data.id !== args.id) {
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -10,6 +10,7 @@ const {
   findResourcesWithAggregation
 } = require('../database/dbOperations');
 const { checkProvenanceHeader, populateProvenanceTarget } = require('../util/provenanceUtils');
+const { checkSupportedResource } = require('../util/baseUtils');
 
 const logger = loggers.get('default');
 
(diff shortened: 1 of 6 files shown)


src/services/base.service.js
```diff
@@ -216,7 +218,9 @@ const baseUpdate = async (args, { req }, resourceType) => {
   logger.info(`${resourceType} >>> update`);
   checkContentTypeHeader(req.headers);
   const data = req.body;
-
+  if (data.resourceType !== undefined) {
+    checkSupportedResource(data.resourceType);
+  }
```
mgramigna, line 223: Including this `undefined` guard for the update case actually allows us to omit the resourceType from HTTP PUT requests. I checked the spec to see if it's allowed, and I think the sentence "The request body SHALL be a FHIR Resource" is enough to infer that the JSON content must match the structure of a FHIR resource.

All this is to say that I think we should be deleting this if block and always checking the supported resource type. That way we can get an error message that says that you can't use "undefined" as a resourceType, which makes sense. Check out what happens when you send this request body in 2 cases:

```
PUT http://localhost:3000/4_0_1/Patient/123

{
  "id": "123"
}
```

and 

```
POST http://localhost:3000/4_0_1/Patient

{
  "id": "123"
}
```

Second one throws an error, first one doesn't

test/util/baseUtils.test.js
```diff
@@ -0,0 +1,10 @@
+const { checkSupportedResource } = require('../../src/util/baseUtils');
+
+describe('Testing base utility functions', () => {
```
mgramigna, line 3: based on my catch in the other comment, maybe we should add a test to make sure that `checkSupportedResource(undefined)` also throws an error here

mgramigna (changes requested):
Looks great. Love the new transaction bundle functionality. One small change request

mgramigna (approved):

PR #59, opened 2021-12-17
Utils directory refactor

sarahmcdougall:
# Summary
Refactored the `utils` directory by separating files and functions, and modifying function params. 

## New behavior
None :)

## Code changes
Various changes have been made to the `utils` directory, such as

- Changing the params in some of the functions to only the attributes that are relevant to the functions. Previously, some of the functions would take in very large params with irrelevant attributes.
- Created `exportUtils.js` for `retrieveExportUrl()`. This function used to be in `measureOperationsUtils`. In my opinion, this function seemed unrelated to the other functions in the file.
- Created `patientUtils.js` for functions related to obtaining patient data. This was created in an attempt to make the catch-all `bundleUtils` file smaller in scope.
- Renamed `measureOperationsUtils.js` to `validationUtils.js`. All of the functions are now for validating parameters for FHIR operations. Consolidated some of the code with a new function that checks for the presence of unsupported params in the request.

# Testing guidance
Run `npm run test` and check that all the unit tests still pass. Go through each of the files in the `utils` directory and check that the none of the functions need to further be separated. Also check that the params make sense for the functions (for example, that functions that previously took in the entire `req` now take in only the relevant attributes from `req`). Try sending a few requests to the server to make sure that all the functionality has remained intact.

Please let me know if you have any suggestions for further refactoring. I tried further breaking down the `bundleUtils` file, but I ran into circular referencing errors when separating the functions into multiple files since they all seem to reference each other. I am also thinking of breaking down `createAuditEventFromProvenance()` into multiple functions, but not sure if that would make things more complicated.


MegRiley (commented):
Looks like this might need a rebase, but otherwise it looks good to me

mgramigna (approved):

PR #60, opened 2021-12-20
Update README.md

sarahmcdougall:
# Summary
Updated README since we support population measure reports for evaluate-measure now.
## Code changes

# Testing guidance


natjoe4 (approved):
Incredible work

PR #61, opened 2021-12-21
Bee queue

natjoe4:
# Summary
Added Bee-Queue to handle bulk import requests

## New behavior
When the server is started, a child process is spun up which creates an instance of Bee-Queue to process $bulk-import jobs. Whenever the client requests a bulk import, a job is pushed to a Redis DB where it can be accessed by the Bee-Queue in a separate process and completed.

## Code changes
- Installed and set up Bee-Queue both for pushing jobs and processing
- Updated index.js to start a child process which runs the processing queue upon server start
- Converted import workflow to outsource processing to child process using Bee-Queue
- Updated testing to close queue upon test end

# Testing guidance
- First, install Redis using `brew install redis`
- Run `brew services start redis`, then check if Redis is running using `redis-cli ping`. More info on this in the README.
Functionality should be unchanged, so following the steps in these two PR's should be sufficient for testing [#54](https://github.com/projecttacoma/deqm-test-server/pull/54) [#34](https://github.com/projecttacoma/deqm-test-server/pull/34), however it is no longer necessary to use the `ndjson-retrieval` branch of bulk-data-utilities. Master will be fine.


src/index.js
```diff
@@ -12,6 +13,8 @@ const config = buildConfig();
 const server = initialize(config, app);
 const logger = loggers.get('default');
 
+childProcess.fork('./src/server/bulkDataProcessor.js');
+
```
natjoe4, line 17: Do we need to add something here that will kill the child process upon exiting using ctrl+c? I messed around with some options but couldn't get anything to work, and never had issues without adding code here.

mgramigna: nah

mgramigna (commented):
This looks good to me. My only suggestion would be to change all the `console.log`s and the `console.warn` to use the built in logger that we have for Asymmetrik to keep to logs consistent.

mgramigna (approved):
jk

PR #64, opened 2021-12-26
Export url fix

sarahmcdougall:
# Summary
This PR changes all instances of `exportURL` with `exportUrl` so that we are properly adhering to the spec. It also adds type filters to the export url returned from `retrieveExportUrl()`.

## New behavior
When `retrieveExportUrl()` is called, the result includes the export Url _and_ a comma-delimited string of types that were included in the parameter array. This only occurs if types were included in the parameter array.

## Code changes
Instances of `exportURL` were changes to `exportUrl` in various files. Additional code was added to `retrieveExportUrl()` to filter out the types in the parameter array and create a comma-delimited string out of them. This string is then concatenated to the exportUrl, which is returned from the function. A unit test was added to test out this functionality

# Testing guidance
Run the unit tests and make sure they pass. You can test this on your own by POSTing a parameters object to `http://localhost:3000/4_0_1/$import` that includes an export Url string. Try sending parameters objects with and without types in the parameter array.

Note this branch is branched off the [utils directory refactor](https://github.com/projecttacoma/deqm-test-server/pull/59)

### DIFF
--- a/src/util/patientUtils.js
+++ b/src/util/patientUtils.js
@@ -8,12 +8,17 @@ const { mapResourcesToCollectionBundle, mapArrayToSearchSetBundle } = require('.
 
 /**
 <<<<<<< HEAD
+<<<<<<< HEAD
  * Wrapper function to get patient data and data
  * requirements for given patient id and map the resources to a collection bundle.
 =======
  * Wrapper function to get patient data for a given patient id and its data
  * requirements and map the resources to a collection bundle.
 >>>>>>> 75d7893 (created patientUtils from bundleUtils)
+=======
+ * Wrapper function to get patient data and data
+ * requirements for given patient id and map the resources to a collection bundle.
+>>>>>>> a878b42 (misc changes)
  * @param {string} patientId patient ID of interest
  * @param {Array} dataRequirements data requirements array obtained from fqm execution
  * @returns {Object} patient bundle as a collection bundle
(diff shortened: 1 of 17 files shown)


src/util/patientUtils.js
```diff
@@ -7,8 +7,18 @@ const { findResourceById, findResourcesWithQuery } = require('../database/dbOper
...
  * Wrapper function to get patient data and data
  * requirements for given patient id and map the resources to a collection bundle.
+=======
+ * Wrapper function to get patient data for a given patient id and its data
+ * requirements and map the resources to a collection bundle.
+>>>>>>> 75d7893 (created patientUtils from bundleUtils)
+=======
+ * Wrapper function to get patient data and data
+ * requirements for given patient id and map the resources to a collection bundle.
+>>>>>>> a878b42 (misc changes)
```
mgramigna, line 21: thoughts?

src/util/patientUtils.js
```diff
@@ -23,7 +33,7 @@ async function getPatientDataCollectionBundle(patientId, dataRequirements) {
  * the resources to a searchset bundle (used for Patient/$everything when
  * we are not concerned with a specific measure)
  * @param {string} patientId patient ID of interest
- * @param {string} base_version base version froma rgs passed in through client request
+ * @param {string} base_version base version from args passed in through client request
```
mgramigna, line 36: nice

src/util/patientUtils.js
sarahmcdougall: I really want to get the point across that this is a wrapper function - if it's not clear I can add a fourth copy of the doc string

mgramigna: :)

mgramigna (approved):

PR #67, opened 2021-12-30
Initial operation outcome

lmd59:
# Summary
Adds an operation outcome for the happy case of a successfully complete bulk import. The operation outcome is saved to a file in `tmp` and then can be served from a `/:base_version/file/:clientId/:fileName` endpoint (which is provided as a url in bulkstatus response).

## New behavior
When the client requests a bulk status that is completed, an operation outcome is created (in database and on filesystem) and the url to that operation outcome file is provided as part of the bulk status response.
If a client requests a file at `/:base_version/file/:clientId/:fileName`, that file will be provided as part of the text of the response.

## Code changes
Updates to bulkstatus service to support creation of the operation outcome. Updates to server, clientfile controller, and clientfile service to support new route and serving up OperationOutcome files in `tmp`. Updates to environment to support `HOST` and `PORT`.

# Testing guidance
Test with `npm run test` to run all tests. Run the server, mongo, redis, and the export server. Use your favorite HTTP request tool to send an import request `http://localhost:3000/4_0_1/$import` (ping me if you need a simple body to send - you can use something like `http://localhost:3001/$export` for the exportURL if you're working with the local export server). In the Content-Location response header should be a slug to the bulk status. You can use this to get thebulkstatus from the server, which will respond with a body containing the outcome url. You can use the outcome url to request the successful OperationOutcome file from the server. The OperationOutcome should be type informational and indicate that "Bulk import successfully completed".


natjoe4 (approved):
This looks good to me! Nice work!

PR #68, opened 2021-12-30
Update node-fhir-server-core to 2.2.7

mgramigna:
no new behavior, just updating the dependency to the new rebased copy of fhir-server-core (published as a patch version)

Just test all functionality and make sure nothing broke.


natjoe4 (approved):
Looks good to me!

PR #69, opened 2022-01-03
Use env vars for redis conn

mgramigna:
# Summary

Add configuration to Redis connection to get bee-queue to work in Docker. `docker-compose up --build` on `main` will fail due to Redis connection refused

## New behavior

BQ connections no longer fail when server is run with docker compose

## Code changes

* Added `REDIS_HOST` and `REDIS_PORT` to `.env`
* Use above vars as connection options to redis when provided
* Update docker compose to use a redis image and configure proper environment for server

# Testing guidance

Ensure bulk jobs can be processed and redis connections are good for both `npm start` case and `docker-compose up --build` case


natjoe4 (approved):
Looks good and tests well for me!

PR #70, opened 2022-01-05
Txn bundle updates

MegRiley:
# Summary

## Code changes

# Testing guidance


test/services/measure.service.test.js
```diff
@@ -157,7 +157,7 @@ describe('testing custom measure operation', () => {
       });
   });
 
-  test('$submit-data uploads txn bundle for valid parameters request', async () => {
+  test.only('$submit-data uploads txn bundle for valid parameters request', async () => {
```
natjoe4, line 160:
```suggestion
  test('$submit-data uploads txn bundle for valid parameters request', async () => {
```

test/services/measure.service.test.js
```diff
@@ -166,6 +166,7 @@ describe('testing custom measure operation', () => {
       .set('x-provenance', JSON.stringify(SINGLE_AGENT_PROVENANCE))
       .expect(200)
       .then(async response => {
+        console.log(JSON.stringify(response.body, null, 4));
```
natjoe4, line 169:
```suggestion
```

src/services/bundle.service.js
```diff
@@ -161,16 +160,36 @@ async function uploadResourcesFromBundle(entries, headers, baseUrl, baseVersion,
...
+
+    return replaceAxiosWithMongo(entry, method).catch(e => {
       return e.response;
     });
   });
   const requestResults = await Promise.all(requestsArray);
   return requestResults;
 }
 
+async function replaceAxiosWithMongo(entry, method) {
```
natjoe4, line 173: I would change this function name to something like `insertBundleResources` or something. I think replaceAxiosWithMongo will be confusing outside of this task.

src/services/bundle.service.js
```diff
@@ -33,17 +34,18 @@ const makeTransactionResponseBundle = (results, res, baseVersion, type, xprovena
     const entry = new Bundle({ response: { status: `${result.status} ${result.statusText}` } });
     if (result.status === 200 || result.status === 201) {
       if (xprovenanceIncluded) {
-        bundleProvenanceTarget.push(JSON.parse(result.headers['x-provenance']).target);
+        bundleProvenanceTarget.push({ reference: `${result.resource.resourceType}/${result.resource.id}` });
       }
-      entry.response.location = result.headers.location;
+      console.log(result.resource.resourceType);
```
natjoe4, line 39:
```suggestion
      
```

src/services/bundle.service.js
```diff
@@ -153,24 +157,67 @@ async function uploadTransactionBundle(req, res) {
...
 }
 
-async function uploadResourcesFromBundle(entries, headers, baseUrl, baseVersion, protocol, xprovenanceIncluded) {
+async function uploadResourcesFromBundle(entries, headers, xprovenanceIncluded, baseVersion) {
   const scrubbedEntries = replaceReferences(entries);
   // define headers to be included in axios call
   const entryHeaders = { 'Content-Type': 'application/json+fhir' };
   if (xprovenanceIncluded) {
     entryHeaders['X-Provenance'] = headers['x-provenance'];
   }
```
natjoe4, line 166: The above lines can be deleted. They were only relevant to the axios request method.

src/services/bundle.service.js
```diff
@@ -153,24 +157,67 @@ async function uploadTransactionBundle(req, res) {
...
+        data: results.toJSON()
+      };
     });
   });
   const requestResults = await Promise.all(requestsArray);
   return requestResults;
 }
 
+async function insertBundleResources(entry, method) {
+  //need to return an array of promises
```
natjoe4, line 188:
```suggestion

```

src/services/bundle.service.js
```diff
@@ -153,24 +157,67 @@ async function uploadTransactionBundle(req, res) {
...
+  if (method === 'PUT') {
+    const { id, created } = await updateResource(entry.resource.id, entry.resource, entry.resource.resourceType);
+    if (created === true) {
+      entry.status = 201;
+      entry.statusText = 'Created';
+    } else if (id != null && created === false) {
+      entry.status = 200;
+      entry.statusText = 'OK';
+    }
+  } else {
```
natjoe4, line 207: I think it's worth adding a unit test with an invalid request type to hit this branch

src/services/bundle.service.js
```diff
@@ -153,24 +157,82 @@ async function uploadTransactionBundle(req, res) {
...
   });
   const requestResults = await Promise.all(requestsArray);
   return requestResults;
 }
 
+/**
+ * Supports Bundle upload to the server using transaction
+ * @param {Object} entry - an object from the TB to insert to the database
+ * @param {string} method - The method of the request, currently on PUT or POST are supported
+ * @returns {Object} an object  containing the results of a  mongo insertion or update for the
```
natjoe4, line 200:
```suggestion
 * @returns {Object} an object containing the results of a mongo insertion or update for the
```

test/services/bundle.service.test.js
```diff
@@ -58,9 +67,19 @@ describe('Test transaction bundle upload', () => {
...
+  test('error thrown if method  type is not PUT or POST', async () => {
+    await supertest(server.app)
+      .post('/4_0_1/Measure/$submit-data')
+      .send(INVALID_METHOD_REQ)
+      .set('Accept', 'application/json+fhir')
+      .set('content-type', 'application/json+fhir')
+      .set('x-provenance', JSON.stringify(SINGLE_AGENT_PROVENANCE))
+      .expect(400)
+      .then(async response => {
+        expect(response.statusCode).toEqual(400);
```
natjoe4, line 79: So I think we're actually hitting the wrong error case here. I believe this 400 is thrown because there is no resource attribute in the entry of the fixture used. If you add a resource attribute with a valid resourceType and id, then we'll get the expected error case, which should be verified in this test by checking the resulting error message

natjoe4, line 79:
```suggestion
```
We already check the status code with the `.expect(400)` above

package.json
```diff
@@ -27,7 +27,6 @@
     "@asymmetrik/fhir-qb": "^0.12.5",
     "@asymmetrik/fhir-qb-mongo": "^0.12.3",
     "@projecttacoma/node-fhir-server-core": "^2.2.7",
-    "axios": "^0.21.4",
```
mgramigna, line 30: This was a result of the connectathon so no fault here, but figured I'd mention. We actually do still need axios now since the ndjson workers are sending requests to get the ndjson content from the export. You'll see that when you rebase

test/services/bundle.service.test.js
```diff
@@ -58,9 +79,20 @@ describe('Test transaction bundle upload', () => {
       });
   });
 
-  afterAll(cleanUpTest);
+  test('error thrown if method  type is not PUT or POST', async () => {
```
natjoe4, line 82:
```suggestion
  test('error thrown if method type is not PUT or POST', async () => {
```

test/services/bundle.service.test.js
```diff
@@ -58,9 +79,20 @@ describe('Test transaction bundle upload', () => {
...
+    await supertest(server.app)
+      .post('/4_0_1/')
+      .send(INVALID_METHOD_REQ)
+      .set('Accept', 'application/json+fhir')
+      .set('content-type', 'application/json+fhir')
+      .set('x-provenance', JSON.stringify(SINGLE_AGENT_PROVENANCE))
+      .expect(200)
+      .then(async response => {
+        expect(response.body.resourceType).toEqual('Bundle');
+        expect(response.body.entry[0].response.status).toEqual('400 BadRequest');
```
natjoe4, line 92: Because there are a bunch of possible 400 errors here, I think we should also check that response.body.entry[0].response.outcome.issue[0].details.text is equal to the expected error message. Something like "Expected requests of type PUT or POST, received GET for test-measure" in this case

test/services/bundle.service.test.js
```diff
@@ -13,11 +13,32 @@ const server = initialize(config);
...
+  params: { base_version: '4_0_1' },
+  headers: { 'content-type': 'application/json+fhir' }
+};
+const INVALID_METHOD_REQ = {
+  resourceType: 'Bundle',
+  type: 'transaction',
+  entry: [
+    {
+      resource: {
+        resourceType: 'Parameters',
```
natjoe4, line 31: Parameters doesn't come up as an accepted resource for me, so this requests still hits the wrong 400 error
```suggestion
        resourceType: 'Measure',
```

natjoe4 (approved):

PR #71, opened 2022-01-06
Gaps post

natjoe4:
# Summary
Fixed a bug that cause POST requests for $care-gaps operations to fail

## New behavior
- POST requests now work for $care-gaps operations
- Request parameters can now be submitted in the url query, request body or both for POST requests

## Code changes
- Added gatherParams function to validateUtils which gathers the params from both the request body and url query
- Added conditionals in careGaps function to allow the fhirQueryBuilder to properly parse the request query params
- Added testing

# Testing guidance
- Run `npm run db:reset`
- Run `npm run connectathon-upload`
- Run `npm run start`
- Send a `GET` request to `http://localhost:3000/4_0_1/Measure/$care-gaps?measureId=measure-EXM130-7.3.000&periodStart=2019-01-01&periodEnd=2019-12-31&subject=denom-EXM130&status=open` and ensure output matches the same request in main.
- Change the request to a `POST` and resend. Ensure output is the same.
- Add a FHIR parameters body to your `POST` request. Within the parameter array, add objects of the following form:
`{"name": "measureId", "valueString":"measure-EXM130-7.3.000"}` where name contains the parameter name and value[x] contains the parameter value, where x is the type of the parameter (eg. valueDate, valueId, valueString, valueCode)
Ensure that the output of these requests is the same.
- Try mixing parameters between the query and the body and ensure the requests come back the same.

### DIFF
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -24,7 +24,6 @@ const {
   findOneResourceWithQuery,
   findResourcesWithQuery
 } = require('../database/dbOperations');
-const { request } = require('express');
 
 const logger = loggers.get('default');
 
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -10,7 +10,8 @@ const { retrieveExportUrl } = require('../util/exportUtils');
 const {
   validateEvalMeasureParams,
   validateCareGapsParams,
-  validateDataRequirementsParams
+  validateDataRequirementsParams,
+  gatherParams
 } = require('../util/validationUtils');
 const {
   getMeasureBundleFromId,
@@ -23,6 +24,7 @@ const {
   findOneResourceWithQuery,
   findResourcesWithQuery
 } = require('../database/dbOperations');
+const { request } = require('express');
 
 const logger = loggers.get('default');
 
@@ -279,11 +281,21 @@ const evaluateMeasure = async (args, { req }) => {
 const careGaps = async (args, { req }) => {
   logger.info('Measure >>> $care-gaps');
 
-  validateCareGapsParams(req.query);
+  let query;
+  if (req.method === 'POST') {
+    query = gatherParams(req.query, req.body);
+  } else {
+    query = req.query;
+  }
+  validateCareGapsParams(query);
 
-  const { periodStart, periodEnd, subject } = req.query;
-  const searchTerm = retrieveSearchTerm(req);
-  req.query = searchTerm;
+  const { periodStart, periodEnd, subject } = query;
+  const searchTerm = retrieveSearchTerm(query);
+  if (req.method === 'POST') {
+    req.body = searchTerm;
+  } else {
+    req.query = searchTerm;
+  }
   //Use the base search function here to allow search by measureId, measureUrl, and measureIdentifier
   const measure = await search(args, { req });
   if (measure.total === 0) {
@@ -319,8 +331,8 @@ const careGaps = async (args, { req }) => {
  * @param {Object} req http request object
  * @returns {Object} an object containing the measure identifier with the appropriate key
  */
-const retrieveSearchTerm = req => {
-  const { measureId, measureIdentifier, measureUrl } = req.query;
+const retrieveSearchTerm = query => {
+  const { measureId, measureIdentifier, measureUrl } = query;
 
   if (measureId) {
     return { _id: measureId };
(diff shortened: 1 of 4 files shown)


src/util/validationUtils.js
```diff
@@ -169,9 +169,30 @@ const checkNoUnsupportedParams = (query, unsupportedParams, operationName) => {
...
+ * @param {Object} body a FHIR Parameters object passed into the body of the request passed in
+ * @returns {Object} an object containing a combination of request parameters from both sources
+ */
+const gatherParams = (query, body) => {
+  const params = { ...query };
+
+  if (body.parameter) {
+    body.parameter.reduce((acc, e) => {
+      if (!e.resource) {
+        acc[e.name] = e.valueDate || e.valueString || e.valueId || e.valueCode;
```
natjoe4, line 184: Is there a better way to accomplish this? Not sure how to specify the expected key for given params here.

src/util/validationUtils.js
```diff
@@ -169,9 +169,30 @@ const checkNoUnsupportedParams = (query, unsupportedParams, operationName) => {
   }
 };
 
+/**
+ * Pulls query parameters from both the body and creates a new parameters map
+ * @param {Object} query the query terms on the request URL
+ * @param {Object} body a FHIR Parameters object passed into the body of the request passed in
```
sarahmcdougall, line 175:
```suggestion
 * @param {Object} body http request body
```

sarahmcdougall, line 175: I'm pretty sure we are bringing in the entire request body (which isn't really an issue because it's expected to just be a FHIR Parameters resource), so we should take note that this is technically the entire request body

src/util/validationUtils.js
sarahmcdougall: I think that for now this approach is good since these are the only value[x] keys that need to be supported for the care gaps Parameters object, although that may restrict this function to only being valid for the care gaps operation. In the future, maybe we can add a function to determine what value[x] keys we need to support for the given operation, and then we can pass those value[x] keys into the `gatherParams()` function. I think this would be out of scope for now.

sarahmcdougall: I do think that we should add a comment above this line explaining that these are the only keys that we need to support for now

src/services/measure.service.js
```diff
@@ -319,8 +330,8 @@ const careGaps = async (args, { req }) => {
  * @param {Object} req http request object
```
sarahmcdougall, line 330:
```suggestion
 * @param {Object} query http request query
```

src/services/measure.service.js
```diff
@@ -279,11 +280,21 @@ const evaluateMeasure = async (args, { req }) => {
 const careGaps = async (args, { req }) => {
   logger.info('Measure >>> $care-gaps');
 
-  validateCareGapsParams(req.query);
+  let query;
+  if (req.method === 'POST') {
+    query = gatherParams(req.query, req.body);
```
sarahmcdougall, line 285: The `gatherParams()` function name isn't super descriptive but I don't necessarily think we should change the function name because otherwise it will be super long. Maybe we should add a comment here explaining that gatherParams creates a new object of the parameters from both the request query and request body

sarahmcdougall (changes requested):
Nice work figuring out how to incorporate both sources of params into the care gaps operation! The functionality is all there, and the unit tests pass. Overall, the code additions are clean and efficient. I added a few simple suggestions/comments to reflect the changes that were made.

Do we want to add anything to the README under the care gaps section to note that we support POST requests and that parameters can be given both in the url query and in the request body? On the other hand, POST is apparently required anyway, so maybe we don't need to explicitly note this in the README. 

I tested the new functionality by sending a FHIR Parameters resource containing a parameter entry with just the name but no valueString, and then with the valueString but no name. In either case, we get an error that we are missing a required parameter, which is good. I'm not sure if we would want to include additional error messages stating that either the name/value string is missing. I think that if the name is missing and you just have a value, then it's okay to just say you are missing a required parameter. On the other hand, if you have the name but not the value, the user might be confused as to how they are "missing" the parameter when they have specified the name. In that case, maybe there should be an error that a parameter name was supplied but no value was provided? Maybe I'm not trusting the user enough to send appropriate request bodies. Up to you if you think the current error message is good or if we should expand upon it!

sarahmcdougall (approved):

PR #72, opened 2022-01-12
Added a count of the # of bulk imports completed

MegRiley:
# Summary
 Added a counter of the number of imports completed and added storing  the counter  as part of the bulk status object

## New behavior
The behavior should be unchanged to the user however there is a now a counter that can be added to the operation outcomes/queries about bulk status to inform a user how many of the imports have successfully completed. 

## Code changes
 src/database/dbOperations.js - modified `completeBulkImportRequest` to now store a counter of bulk imports done as a property of bulk status

src/server/importWorker.js  - modified  `executePingAndPull` to add a counter and increment it when it uploads the transaction bundles. Also added the passing of this counter to the db to store it for future use. 

# Testing guidance
 Ran unit tests, assured they passed. 
Then tried to import an ndjson file I had created with 5 lines. modified the code to log out the value when the import was complete and ensured it was 5 as I had expected. (Note to self should probably add a unit test for this)


lmd59 (commented):
Is it worth creating a new or adding to a test to check this functionality?

src/services/bulkstatus.service.js
```diff
@@ -91,7 +91,7 @@ async function checkBulkStatus(req, res) {
               display: 'New resource created'
             }
           ],
-          text: 'Bulk import successfully completed'
+          text: `Bulk import successfully completed, successfully imported ${bulkStatus.exportedResourceCount}`
```
mgramigna, line 94: Technically, `exportedResourceCount` is not always guaranteed to be here. You can see in the code that we have a few checks where it can be `-1`.

This value is computed by looking at the `count` property returned by the initial `$export` request as part of PnP. The export server does not have to specify this count, though they can if they want to. So this code will break if we are dealing with an export server that doesn't include `count` on the exported ndjson, and it will say successfully imported -1.

What we should be doing is grabbing the count of resources actually inserted by the ndjson worker job and tracking that information. We can check the number of promises resolved by inspecting how many results came back from `Promise.all`. Then we can use that to update the bulkStatus information in the db. We shouldn't be relying on `bulkStatus.exportedResourceCount`

For future work with the error handling stuff too, it's important that we get the count in the ndjson worker itself, because we want to make sure that we get a true value on how many mongo insertions were successfully done. What this current code is doing is saying how many resources we _expected_ to have inserted, not actually inserted

src/server/ndjsonWorker.js
```diff
@@ -41,8 +45,16 @@ ndjsonWorker.process(async job => {
...
     });
+  let count = await getCurrentSuccessfulImportCount(clientId);
+  const results = await Promise.all(insertions);
 
-  await Promise.all(insertions);
+  results.map(result => {
+    if (result.created === true) {
+      count = count + 1;
+    }
+  });
```
mgramigna, line 55: This block of code will not work in the case of importing resources that already exist.

You shouldn't have to iterate over the results, and you can just use `results.length` as the number of successful insertions.

This loop results in count being `0` every time if we repeat the same import call multiple times since it will be doing `update` instead of `create` (created is always false then)

src/server/ndjsonWorker.js
```diff
@@ -41,8 +45,16 @@ ndjsonWorker.process(async job => {
...
+  console.log('the value of count is: ' + count);
+  await updateSuccessfulImportCount(clientId, count);
```
mgramigna, line 57: This will run into a weird race condition where multiple ndjson workers could be updating the count at one time. 

The strategy we had around this was to use `process.send` as you can see a few lines below this in order to tell the main process to update the db from there. We can add the number of successfully resolved promises to the object in the `process.send` call to pass along the total number of resources to update the count with. Then you can do this database update from there.

src/services/bulkstatus.service.js
```diff
@@ -91,7 +91,7 @@ async function checkBulkStatus(req, res) {
               display: 'New resource created'
             }
           ],
-          text: 'Bulk import successfully completed'
+          text: `Bulk import successfully completed, successfully imported ${bulkStatus.successCount}`
```
mgramigna, line 94:
```suggestion
          text: `Bulk import successfully completed, successfully imported ${bulkStatus.successCount} resources`
```

mgramigna (changes requested):
Seeing bugs caused by inline comments. Should hopefully be simple enough changes.

I also noticed that this branch doesn't appear to be properly rebased. We have some commits here that have already been merged into master and the code diff is showing erroneous changes that you didn't actually include with this task (e.g. the diagram for bulk submit data). Is there any way we can update this branch such that it only shows the relevant code diff for just this functionality?

src/server/ndjsonWorker.js
```diff
@@ -41,12 +41,14 @@ ndjsonWorker.process(async job => {
       const data = JSON.parse(resourceStr);
       return updateResource(data.id, data, data.resourceType);
     });
+  let sucessCount = await getCurrentSuccessfulImportCount(clientId);
```
lmd59, line 44: success is spelled incorrectly here and below, which might be a bit confusing

src/server/importWorker.js
```diff
@@ -25,6 +25,7 @@ importQueue.process(async job => {
   // Call the existing export ndjson function that writes the files
   console.log(`import-worker-${process.pid}: Kicking off export request: ${exportURL}`);
   const result = await executePingAndPull(clientEntry, exportURL, measureBundle);
+
```
lmd59, line 28: Def a nitpick, but maybe we can get rid of these line changes since there aren't any substantive changes to the importWorker?

lmd59 (commented):
I haven't tested this out or been very thorough yet, but here are some initial comments.

package-lock.json
```diff
@@ -2662,14 +2662,15 @@
       "dev": true
     },
     "bulk-data-utilities": {
-      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#92831b38717ee8ef1350d57a06423c09092ed7fa",
+      "version": "git+https://github.com/projecttacoma/bulk-data-utilities.git#0fc2f3dadba664fadeba2b285bcc30362cb3a043",
```
mgramigna, line 2665: Not sure how this happened, but this actually downgrades the bulk-data-utilities dependency (which could be why you're seeing issues with bulk data stuff that others aren't). the `92831b` commit is the proper hash we should be using

PR #73, opened 2022-01-13
Jan 2022 Connectathon Changes

sarahmcdougall:
# Summary
We changed all the things.

## New behavior

**New NDJSON Worker:**
Previously, we only had child processes running with import workers. Now, we also have child processes running for NDJSON workers. The NDJSON worker pulls down a job from the new NDJSON queue and retrieves the NDJSON from the file URL specified in the job. It then connects to our Mongo database and inserts the NDJSON resource. **(RIP BDU sqlite 2021-2022 :broken_heart:)**

**Keeping track of bulk files to process:**
Now, the import request completes once we have exported all the files, and we keep track of this by initializing the bulk file count and then decrementing by one after each file finishes exporting.

We decided to use this approach for now because the database updates need to happen from the main process to avoid race conditions. So far, we have not found a solution that is more efficient than our approach of decrementing the bulk file count. In the future, we can look more into a lockfile approach for handling this functionality. 

**Dynamic X-Progress Header:**
For import requests that are in progress, the X-Progress header will now dynamically update with the percentage of exported files or resources. Since number of resources is more accurate, we will use the resource counts if they are available by the export server. Otherwise, we resort to the number of files that have been exported thus far.

## Code changes

**`executePingAndPull() changes`:**
We got rid of the transaction bundle uploading in `executePingAndPull()`. Instead, we make use of the bulk file count approach with the child processes, enqueuing a parsing job onto our NDJSON queue for each NDJSON file.

**Dynamic X-Progress Header:**
The bulk import status entries in Mongo now contain attributes to keep track of the exported file/resource counts, and total file/resource counts. If the resource counts are not available, these entries will default to -1. The X-Progress header is computed within `bulkstatus.service`.

**ValueString —> ValueURL:**
After discussions on Zulip, we concluded it is more appropriate to use `ValueUrl` for the `exportUrl` rather than `ValueString`, which is what we had been using. We updated all instances of `ValueString` in `src`, unit tests, and fixtures.

**`exportType ServerError`:**
To align more with the spec, we added a NotImplemented ServerError that is thrown when an import request is sent that contains a `”static” exportType`. We only support `”dynamic” exportType`s.

**Full URL in `Content-Location`:**
We had some issues getting the [SMART on FHIR Bulk Import Client](https://github.com/smart-on-fhir/bulk-import-client) to work with the test server because we were not setting the `Content-Location` to its full URL. We made changes to `measure.service` and `import.service` to populate the header with the full URL.

# Testing guidance
Run all unit tests and ensure they pass.

Try sending some $import requests within the test server. For example, try sending a request to `http://localhost:3000/4_0_1/$import` with the following request body:

```
{
	"resourceType": "Parameters",
	"parameter": [
		{
			"name": "exportUrl",
			"valueUrl": "https://bulk-data.smarthealthit.org/eyJlcnIiOiIiLCJwYWdlIjoxMDAwMCwiZHVyIjoxMCwidGx0IjoxNSwibSI6MSwic3R1IjozLCJkZWwiOjB9/fhir/$export"
		},
		{
			"name": "exportType",
			"valueString": "dynamic"
		}
		
	]
}
```

After sending the GET request to the specified `Content-Location`, make sure to check the headers and see the super cool dynamic X-Progress header, and admire how fast everything is. Also test out $import with our bulk export server to ensure functionality is as expected. (Don’t forget to add `/$export` to the end of the `valueUrl`). 

Note: Once the BDU PR gets approved, we will need to update the commit hash for bulk-data-utilities in package.json, as the current commit hash points to the BDU Connectathon Branch.


lmd59:
I'm not seeing any BDU PRs since 11 days ago. Does the BDU PR still need to go up?

lmd59 (approved):
I dunno man- couldn't find anything that I thought was wrong. Looks good to me!

PR #75, opened 2022-01-20
Connectathon Unit Tests

sarahmcdougall:
# Summary
Unit tests were added to the test server to reflect code changes from the Connectathon.

## New behavior
Behavior should remain unchanged.

## Code changes
Unit tests were added for the following scenarios:
- Checks that the bulk file counts and bulk resource counts are properly initialized and properly decremented. Checks that the bulk status becomes "complete" once the bulk file count reaches 0.
- Checks that the X-Progress header is populated with the % complete. A test was added that uses file counts, and a test was added that uses resource counts.
- In validationUtils, checks that an error is thrown when a "static" exportType is included in the parameters array.
- Miscellaneous tests were added to bring up the overall test coverage.

# Testing guidance
Run `npm run test` and ensure that all the unit tests pass. Please check the unit tests to make sure their logic is correct.

Run `npm run test:coverage` to view the test coverage for all statements, branches, functions, and lines. Currently, everything is > 80%. 

Please let me know if you think any of the uncovered lines should be covered.

### DIFF
--- a/src/util/validationUtils.js
+++ b/src/util/validationUtils.js
@@ -193,7 +193,7 @@ const gatherParams = (query, body) => {
 /**
  * Uses request body parameter to search for the optional exportType parameter if present.
  * If so, checks that exportType is not static
- * @param {Object} parameters - request body parameter
+ * @param {Array} parameters - parameters array from request body
  */
 const checkExportType = parameters => {
   const exportTypes = parameters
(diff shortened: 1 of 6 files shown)


natjoe4 (changes requested):
Coverage looks great and all tests work as advertised. Only change is that, for the tests in dbOperations, the expect statements don't need to be in a .then() function since you already await the BulkFileCount function. I think it's more intuitive to remove the .then() function and just run the lines.

natjoe4 (approved):

PR #76, opened 2022-01-21
added eCQM-specific bulk import diagram

natjoe4:
# Summary
Added a diagram which will be used in the eCQM-specific bulk import section of the wiki
## New behavior
none
## Code changes
none
# Testing guidance
shouldn't be necessary, but take a look at the chart :)


sarahmcdougall (approved):
Looks like this branch might need a rebase but otherwise looks good and the changes I requested last week were made :)

PR #79, opened 2022-01-27
OperationOutcome file for individual resource errors during bulk import

natjoe4:
# Summary
Added a new outcome to bulk Import response with access to an error NDJson file when appropriate

## New behavior
When some or all resources fail during an $import or bulk submit data request, the bulkstatus endpoint will now provide a second OperationOutcome ndjson URL. Sending a `GET` request to this URL will return an NDJson file of OperationOutcome resources, one for each resource that failed import, with information on why the resources failed

## Code changes
- Added new failedOutcomes array to bulkstatus entry which stores error messages
- Added new function in dbOperations to push to this failedOutcomes array
- Added functionality to bulkstatus service that constructs the ndjson file on bulkstatus query and returns a URL at which the ndjson file can be accessed
- Added error for situation where bulkExportServer exports no resources (previously caused the request to hang)
- Added testing and updated existing fixtures

# Testing guidance
- Run the new tests using `npm test`
- Run a typical $import and bulk submit data request and ensure that the output is as expected
- Add the line `resourceStr = resourceStr.slice(1)` on line 44 in src/server/ndjsonWorker.js. This will cause the JSON.parse command to fail.
- Run a typical $import and bulk submit data request, follow the bulkstatus endpoint and ensure that the Error.ndjson url appears
- Send a `GET` request to the Error.ndjson url and ensure that the number of resources that failed on the parsing step is equivalent to the total number of resources that the bulk export server exported (They should all fail)
- Delete the JSON scrambling line
- Add the line `data.resourceType = 'INVALID'` on line 49 in src/server/ndjsonWorker.js. This will cause the `checkSupportedResourceType()` call to fail
- Repeat the $import and bulk submit steps detailed above and confirm the same results, this time with different error messages
- NOTE: Take special care to note the number of errors given and compare it to the number expected. It's unclear if Mongo db can handle the potential race conditions caused by pushing to a stored array from multiple processes.

### DIFF
--- a/src/services/bulkstatus.service.js
+++ b/src/services/bulkstatus.service.js
@@ -98,7 +98,7 @@ async function checkBulkStatus(req, res) {
     const OperationOutcome = resolveSchema(req.params.base_version, 'operationoutcome');
     writeToFile(JSON.parse(JSON.stringify(new OperationOutcome(outcome).toJSON())), 'OperationOutcome', clientId);
 
-    return {
+    const response = {
       transactionTime: '2021-01-01T00:00:00Z',
       requiresAccessToken: true,
       outcome: [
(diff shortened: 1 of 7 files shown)


src/services/bulkstatus.service.js
```diff
@@ -109,6 +109,28 @@ async function checkBulkStatus(req, res) {
...
     };
+
+    if (bulkStatus.failedOutcomes.length > 0) {
+      bulkStatus.failedOutcomes.forEach(fail => {
+        const failOutcome = {};
+        failOutcome.id = uuidv4();
+        failOutcome.issue = [
+          {
+            severity: 'error',
+            code: 'BadRequest',
```
natjoe4, line 120: Fine to force this to a BadRequest here? I think this can only happen from bad input from the export server, but not sure if that qualifies as a bad request

sarahmcdougall (changes requested):
I tried multiple test cases and they all gave me the correct result! Also tested out the case for when the export server does not export anything, and I got the appropriate error. I just have a few spelling changes but no major code changes. I'll hold off on merging in case we want to get other PRs merged first. Good work!

sarahmcdougall (approved):

PR #80, opened 2022-01-28
$care-gaps returns Parameters object

natjoe4:
# Summary
Converted `$care-gaps` output to FHIR Parameters object instead of Bundle, per the spec

## New behavior
The `$care-gaps` operation will now return a Parameters object upon successful completion

## Code changes
Moved the results of the FQM-execution gaps in care call into the body of a Parameters object before returning in measure.service.js

# Testing guidance

- run `npm run db:reset`
- run `npm run connectathon-upload` after cloning connectathon repository
- send a `POST` request to `http://localhost:3000/4_0_1/Measure/$care-gaps?periodStart=2019-01-01&periodEnd=2019-12-31&subject=denom-EXM130&status=open` with the following body:
```
{
	"resourceType": "Parameters",
	"parameter":[{
		"name":"measureId",
		"valueId":"measure-EXM130-7.3.000"
	}]
}```

- Ensure that the results are stored correctly within a Parameters object and the bundle is unchanged from what is returned from the same request in main.

### DIFF
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -1,4 +1,4 @@
-const { ServerError, loggers, resolveSchema } = require('@projecttacoma/node-fhir-server-core');
+const { ServerError, loggers } = require('@projecttacoma/node-fhir-server-core');
 const { Calculator } = require('fqm-execution');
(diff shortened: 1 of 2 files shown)


mgramigna (approved):
Inspirational

PR #81, opened 2022-01-31
Bringing the code over from #72

MegRiley:
Summary
Added a counter of the number of imports completed and added storing the counter as part of the bulk status object

New behavior
The behavior should be unchanged to the user however there is a now a counter that can be added to the operation outcomes/queries about bulk status to inform a user how many of the imports have successfully completed.

Code changes
src/database/dbOperations.js - added getCurrentSuccessfulImportCount, and updateSuccessfulImportCount to now store and update a counter of bulk imports done as a property of bulk status

src/server/importWorker.js - modified executePingAndPull to add a counter and increment it when it uploads the transaction bundles. Also added the passing of this counter to back to `index.js` so that the value can be written to the database without running into any race conditions

src/services/bulkstatus.service.js - updated the text of the operation outcome to now echo out the counter of successful imports that occurred

Testing guidance
Ran unit tests, assured they passed.
Using the bulk export server, I populated the database with the connectathon data script.
I send a bulk export request modelled after the example in the wiki and ensured I got the 202. Using the location header value I sent a bulk status request. (mostly to ensure no errors, I know you can skip this step and go straight to the 3rd request.) 
Once the $bulkstatus request returned 200 I sent a get to the  url in the body of the operation outcome and confirmed that the server responded that it completed my bulk import and the message contained the number of items imported (386 in this case)

### DIFF
--- a/src/services/bulkstatus.service.js
+++ b/src/services/bulkstatus.service.js
@@ -91,7 +91,7 @@ async function checkBulkStatus(req, res) {
               display: 'New resource created'
             }
           ],
-          text: 'Bulk import successfully completed'
+          text: `Bulk import successfully completed, successfully imported ${bulkStatus.successCount} resources`
         }
       }
     ];
--- a/test/fixtures/testBulkStatus.json
+++ b/test/fixtures/testBulkStatus.json
@@ -41,6 +41,7 @@
     "exportedFileCount": 10,
     "totalResourceCount": 500,
     "exportedResourceCount": 200,
+    "succesCount":200,
     "error": { "code": null, "message": null }
   }
 ]
--- a/test/fixtures/testBulkStatus.json
+++ b/test/fixtures/testBulkStatus.json
@@ -57,6 +57,7 @@
     "totalResourceCount": 500,
     "exportedResourceCount": 200,
     "succesCount":200,
-    "error": { "code": null, "message": null }
+    "error": { "code": null, "message": null },
+    "failedOutcomes": []
   }
 ]
--- a/test/fixtures/testBulkStatus.json
+++ b/test/fixtures/testBulkStatus.json
@@ -33,5 +33,14 @@
     "status": "Error",
     "error": { "code": "ErrorCode", "message": "Known Error Occurred!" }
   },
-  { "id": "UNKNOWN_ERROR_REQUEST", "status": "Error", "error": { "code": null, "message": null } }
+  { "id": "UNKNOWN_ERROR_REQUEST", "status": "Error", "error": { "code": null, "message": null } },
+  {
+    "id": "COMPLETED_REQUEST_WITH_RESOURCE_COUNT",
+    "status": "Completed",
+    "totalFileCount": 100,
+    "exportedFileCount": 10,
+    "totalResourceCount": 500,
+    "exportedResourceCount": 200,
+    "error": { "code": null, "message": null }
+  }
 ]
--- a/test/services/bulkstatus.service.test.js
+++ b/test/services/bulkstatus.service.test.js
@@ -69,6 +69,19 @@ describe('checkBulkStatus logic', () => {
         expect(response.body.issue[0].details.text).toEqual('Could not find bulk import request with id: INVALID_ID');
       });
   });
+  test('check operationOutcome includes the number of resources when available', async () => {
+    await supertest(server.app).get('/4_0_1/bulkstatus/COMPLETED_REQUEST_WITH_RESOURCE_COUNT').expect(200);
+    const response = await supertest(server.app).get(
+      '/4_0_1/file/COMPLETED_REQUEST_WITH_RESOURCE_COUNT/OperationOutcome.ndjson'
+    );
+    response.text
+      .trim()
+      .split(/\n/)
+      .map(async resourceStr => {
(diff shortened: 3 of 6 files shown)


test/services/bulkstatus.service.test.js
```diff
@@ -81,6 +81,13 @@ describe('Dynamic X-Progress logic', () => {
         expect(response.headers['x-progress']).toEqual('60.00% Done');
       });
   });
-
+  test('check operationOutcome includes the number of resources when available', async () => {
+    await supertest(server.app)
+      .get('/4_0_1/bulkstatus/COMPLETED_REQUEST_WITH_RESOURCE_COUNT')
+      .expect(200)
+      .then(response => {
+        expect(response.body.outcome[0].type).toEqual('OperationOutcome');
```
lmd59, line 89: This doesn't seem to be checking that there's some count of resources available? It's just checking the type is OperationOutcome

src/services/bulkstatus.service.js
```diff
@@ -91,7 +91,7 @@ async function checkBulkStatus(req, res) {
               display: 'New resource created'
             }
           ],
-          text: 'Bulk import successfully completed'
+          text: `Bulk import successfully completed, successfully imported ${bulkStatus.successCount} resources`
```
lmd59, line 94: The wording of the test seems to imply that we won't always have a success count. In which case, this sentence will print a bit weird.

lmd59 (commented):
Migrated comments from #72

src/services/bulkstatus.service.js
MegRiley: We should always have a count unless the operation fails completely (then we shouldn't end up here) This count isn't based off the info provided by the export server, it's a count of the imports we were able to do

test/services/bulkstatus.service.test.js
```diff
@@ -81,6 +82,33 @@ describe('Dynamic X-Progress logic', () => {
         expect(response.headers['x-progress']).toEqual('60.00% Done');
       });
   });
+  test('check operationOutcome includes the number of resources when available', async () => {
+    const response = await supertest(server.app)
```
mgramigna, line 86: This is not quite the approach we want to take with the unit tests. What you had in the previous commit was almost it, we just needed to add different assertions and update the test fixtures a bit.

These unit tests should only be testing the bulkstatus endpoint. The reason we have all those bulkstatus test fixtures in that JSON file is so we don't have to go through the import flow within the unit tests. We can just call existing bulkstatus endpoints with known client IDs directly and control what they return. We don't need to call import and grab the content location.

You have a test fixture for a bulkstatus with ID `COMPLETED_REQUEST_WITH_RESOURCE_COUNT`, and you have complete control over what a GET request to `/4_0_1/bulkstatus/COMPLETED_REQUEST_WITH_RESOURCE_COUNT` should return. Namely, whatever JSON exists for that test fixture in the `testBulkStatus.json` file.

Your  code introduces a new property `successCount` on these bulkStatus objects that represents the number of successfully parsed resources, and `bulkStatus.successCount` is what we use to populate the OperationOutcome text with the count. So the unit test changes to get a proper OperationOutcome should involve updating any fixtures and calls to make sure that `successCount` is properly set on the returned bulkStatus. That way, you're testing whether or not the code properly takes that number and puts it on the OperationOutcome resources.

You'll still need to have a supertest request to `/4_0_1/file/COMPLETED_REQUEST_WITH_RESOURCE_COUNT/OperationOucome.ndjson`, but you should be able to hardcode the clientID since it's the same as the bulkstatus request endpoint. You'll still need to look at newResponse.text and make your assertions there, but you need to assert that the proper `successCount` value lives within the OperationOutcome text. This current assertion is just asserting that string "successfully imported  " is contained somewhere within the entire stringified resource, which doesn't quite hit the case that this task is aimed at.

test/fixtures/testBulkStatus.json
```diff
@@ -33,5 +33,14 @@
...
   },
-  { "id": "UNKNOWN_ERROR_REQUEST", "status": "Error", "error": { "code": null, "message": null } }
+  { "id": "UNKNOWN_ERROR_REQUEST", "status": "Error", "error": { "code": null, "message": null } },
+  {
+    "id": "COMPLETED_REQUEST_WITH_RESOURCE_COUNT",
+    "status": "Completed",
+    "totalFileCount": 100,
+    "exportedFileCount": 10,
+    "totalResourceCount": 500,
+    "exportedResourceCount": 200,
```
mgramigna, line 43: See unit test comment for explanation of required changes here

mgramigna (changes requested):

test/services/bulkstatus.service.test.js
```diff
@@ -81,6 +80,19 @@ describe('Dynamic X-Progress logic', () => {
         expect(response.headers['x-progress']).toEqual('60.00% Done');
       });
   });
+  test('check operationOutcome includes the number of resources when available', async () => {
```
mgramigna, line 83: Last comment. This test is currently in a `describe` block for `Dynamic X-Progress logic`, can we move it to the above `describe` block labeled `checkBulkStatus logic`?

mgramigna (changes requested):

mgramigna (approved):

PR #82, opened 2022-02-07
Check for OperationOutcome file before writing and add unit test to c…

lmd59:
…heck for issue

# Summary
Updates to bulk status service to fix a duplicate OperationOutcome issue.

## New behavior
When a user requests a successful bulk status twice and then views the resultant OperationOutcome, there should only be a single OperationOutcome resource.

## Code changes
Updates to bulkstatus service check for an existing OperationOutcome file before writing the OperationOutcome so that it does not get written every time there's a status call.

# Testing guidance
Full steps to replicate issue:
- Send a standard bulk $import request
- Send the status poll request
- (wait to get a 200)
- Send OperationOutcome get request (should be only one OperationOutcome)
- send the status poll request again (should still get 200)
- Send OperationOutcome get request again
- OperationOutcome ndjson contains a duplicate of the previous entry

Updates to bulk status service should fix issue of duplicate OperationOutcome. Run `npm run test` and all tests should pass. New test should fail with previous code, but should pass with the updates to bulkstatus service.

Question...
Would we have a situation where we'd want to add more than one OperationOutcome to the file? We currently only write to file when the status is completed or if the status fails. If we'd want to write multiple to a file, we might need to reorganize this.

### DIFF
--- a/src/services/bulkstatus.service.js
+++ b/src/services/bulkstatus.service.js
@@ -163,6 +163,9 @@ async function checkBulkStatus(req, res) {
 }
 
 const writeToFile = function (doc, type, clientId) {
+  // Do not write if file already has contents
+  if (checkForFile(type, clientId)) return;
+
   const dirpath = './tmp/' + clientId;
   fs.mkdirSync(dirpath, { recursive: true });
   const filename = path.join(dirpath, `${type}.ndjson`);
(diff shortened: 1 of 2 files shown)


sarahmcdougall (approved):
All the unit tests pass. I tested the new code in Insomnia and we no longer run into that duplication issue. No codes suggestions - everything looks good!

Based on my understanding of the bulk import pnp spec, I don't think we will ever run into a situation where we already have the file created and want to append to it. I talked with Matt a bit and he agrees that we will likely not have a case where we want to append to an already-created OperationOutcome file. There may be a time where we want to add multiple OperationOutcomes at the same time to the same file, and so in this case I think we would need to combine all the OperationOutcome JSON to use as the parameter for `writeToFile()` since we will only want to call `writeToFile()` once.

PR #84, opened 2022-02-07
Change open status to open-gap

sarahmcdougall:
# Summary
Updates were made to the care-gaps operation to support `open-gap` status instead of `open` so that our implementation aligns with the spec.

## New behavior
After observing the [care-gaps TestScripts in Touchstone](https://touchstone.aegis.net/touchstone/testdefinitions?selectedTestGrp=/FHIRSandbox/DaVinci/FHIR4-0-1-DEQM-GIC&activeOnly=false&contentEntry=TEST_SCRIPTS ) and some of the examples in the [care-gaps spec](https://build.fhir.org/ig/HL7/davinci-deqm/OperationDefinition-care-gaps.html), I noticed that we are using `open` for status instead of `open-gap`. The test server now supports `open-gap` for the `status` parameter.

## Code changes
Changes were made to the status validation code and the unit tests that use the status parameter.

# Testing guidance
Check that all instances of `open` have been changed to `open-gap`. Run the unit tests and make sure they pass. Try sending a care-gaps request in Insomnia with `status=open`. You should receive a ServerError explaining that we only support `status='open-gap'`. Then, send a care-gaps request with `status=open-gap`. You should receive 200 OK.

### DIFF
--- a/src/util/validationUtils.js
+++ b/src/util/validationUtils.js
@@ -90,7 +90,7 @@ const validateCareGapsParams = query => {
     });
   }
 
-  if (query.status !== 'open') {
+  if (query.status !== 'open-gap') {
     throw new ServerError(null, {
       statusCode: 501,
       issue: [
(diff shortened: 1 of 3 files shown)


natjoe4 (approved):
Looks good to me!

PR #85, opened 2022-02-09
Added Winston Debug and Info Logging

natjoe4:
# Summary
Added debug and info level logging through Winston to clarify the order or operations and assist with debugging.

## New behavior
- Running with NODE_ENV set to anything but 'debug' (defaults to 'production') will print only info-level logs
- Running with NODE_ENV set to 'debug' will print debug-level logs
- ndjsonWorker and importWorker now log using the existing logger instead of `console.log()`
- Logs will no longer print during unit tests

## Code changes
- Added a new .env variable to silence and trigger debug logging
- Added lots and lots of debug logging
- Added high-level info logging
- Added a jest config file to silence logging during unit tests

# Testing guidance
- Ensure all input and output to/from the server is unchanged
- Run using the command `NODE_ENV=development npm start` and ensure all debug-level logs appear
- Run using `npm start` and ensure only info-level logs appear
- Run `npm test` and ensure no logs appear
- Look through all files and determine if all potentially useful info is logged or if any unnecessary info is logged


src/util/validationUtils.js
```diff
@@ -2,7 +2,7 @@ const { ServerError } = require('@projecttacoma/node-fhir-server-core');
 
 /**
  * Checks that the parameters input to $evaluate-measure are valid. Throws a ServerError
- * for missing parameters, the use of unsupported parameters, and the use of unsuppported
+ * for missing parameters, the use of unsupported parameters, and the use of unsupported
```
mgramigna, line 5: And the student becomes the master

natjoe4: vs code spell checker baybeeee

mgramigna (changes requested):
Looks great. Only a few small comments inline.

Note: We definitely take a performance hit in debug mode when we log FHIR libraries since that have huge b64-encoded strings on them (e.g. `$evaluate-measure` takes way longer). I think this performance hit is totally fine since since debug mode is not the default. Just wanted to note it here as long as you agree

mgramigna (approved):

PR #86, opened 2022-02-10
Support for population evaluate-measure for group

lmd59:
# Summary
Adds support for evaluate-measure population report with a group-based subject.

## New behavior
An evaluate-measure request can be sent to the server for evaluation of a group-based population of patients. The created report matches the patients designated by the group. Appropriate errors are created for improperly created subject parameters.

## Code changes

# Testing guidance
`npm run test` all tests should pass. 
Create a Group resource. Example, PUT http://localhost:3000/4_0_1/Group/EXM130-patients with body:
`{
	"resourceType": "Group",
    "id": "EXM130-patients",
    "type": "person",
    "actual": "true",
    "member": [
    {
      "entity": {
        "reference": "Patient/denom-EXM130"
      }
    },
    {
      "entity": {
        "reference": "Patient/numer-EXM130"
      }
    }
  ]
}`
Test evaluate measure using created Group id for the subject. Example, GET http://localhost:3000/4_0_1/Measure/measure-EXM130-7.3.000/$evaluate-measure?subject=Group/EXM130-patients&periodStart=2019-01-01&periodEnd=2019-12-31&reportType=population

Body returned should be the population report for just measures in the group. Consider testing with improper subject/population match, like using `reportType=population` with a patient subject or unavailable group id. Should give appropriate error response.


lmd59:
Other comment from @natjoe4 : now that we accept multiple resource types as valid subjects, could we add a more specific error that tells users that their subject is invalid if the reportType is individual and the subject is Group/Id, or {nonPatientfhirResource}/Id. Something like `For report type 'individual', subject may only be a Patient resource of format "Patient/{id}".`

lmd59:
> Other comment from @natjoe4 : now that we accept multiple resource types as valid subjects, could we add a more specific error that tells users that their subject is invalid if the reportType is individual and the subject is Group/Id, or {nonPatientfhirResource}/Id. Something like `For report type 'individual', subject may only be a Patient resource of format "Patient/{id}".`

I've added this, but note that it doesn't check if an individual (non-reference) id is a patient. Doing a database request for validation seemed like a bit much.

natjoe4 (approved):
Was able to test all desired functionality and code looks good to me!

PR #89, opened 2022-02-18
Group Support for $care-gaps

sarahmcdougall:
# Summary
This PR adds `$care-gaps` Group support and `$care-gaps` subject validation.

## New behavior
Now, a `$care-gaps` request can be sent with a FHIR Group resource as the subject. This case would happen when we have a FHIR Group resource on the server that references some set of patients, and we want to call the `$care-gaps` operation for the patients referenced in that Group resource.

## Code changes
Changes were made in `measure.service.js` to create `patientBundles` for the patients referenced in the Group resource, if the subject specifies a valid Group resource. Otherwise, the patient bundle will be retrieved for the single Patient resource that is specified as the subject.

Updates were made in `validationUtils.js` to check that a resource type is specified in the `subject` parameter, and that the resource type is either `Group` or `Patient`.

Changes were also made in `calculateDataRequirements()` to reflect the requested changes in [TACOSTRAT-878](https://jira.mitre.org/browse/TACOSTRAT-878) with specifying `measurementPeriodStart` and `measurementPeriodEnd`.

# Testing guidance
Check that the new unit tests pass with `npm run test`. The new tests check for `$care-gaps` Group support and that errors are appropriately thrown for an invalid `subject` parameter.

**Note:** This PR updates the dependency for `fqm-execution` to 0.9.0, so make sure to run `npm install` before testing.

To test in Insomnia,
* Create a Group Resource. For example, you can send a PUT request to `http://localhost:3000/4_0_1/Group/EXM130-patients` with the following body:

```json
{ "resourceType": "Group", "id": "EXM130-patients", "type": "person", "actual": "true", "member": [ { "entity": { "reference": "Patient/denom-EXM130" } }, { "entity": { "reference": "Patient/numer-EXM130" } } ] }
```
* Send a GET request to `http://localhost:3000/4_0_1/Measure/$care-gaps?measureId=measure-EXM130-7.3.000&subject=Group/EXM130-patients&periodStart=2019-01-01&periodEnd=2019-12-31&status=open-gap`
* Check that the `parameter` array is of length 2 and that the results of the care gaps operation are included in each entry of the `parameter` array. Note that for the EXM130 numerator patient, the `resource` will be an empty object.
* Test some of the error cases in Insomnia, like including a subject that does not include a resource type, or a subject with an invalid resource type (not Patient or Group)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1962,6 +1962,29 @@
         "path": "^0.12.7",
         "sqlite": "^4.0.23",
         "sqlite3": "^5.0.2"
+      },
+      "dependencies": {
+        "commander": {
+          "version": "6.2.1",
+          "resolved": "https://registry.npmjs.org/commander/-/commander-6.2.1.tgz",
+          "integrity": "sha512-U7VdrJFnJgo4xjrHpTzu0yrHPGImdsmD95ZlgYSEajAn2JKzDhDTPG9kBTefmObL2w/ngeZnilk+OV9CG3d7UA=="
+        },
+        "fqm-execution": {
+          "version": "0.8.1",
+          "resolved": "https://registry.npmjs.org/fqm-execution/-/fqm-execution-0.8.1.tgz",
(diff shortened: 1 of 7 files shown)


lmd59 (approved):
Looks good!

PR #90, opened 2022-02-23
No id gaps

natjoe4:
# Summary
A recent investigation into the $care-gaps IG discovered that the parameters which specify the measure (measureId, measureIdentifier, measureUrl) are all 0...*. Previously, we were throwing an error if a user omitted all three of these measure identification parameters. This PR removes that error, and instead runs $care-gaps for the subject on all measures in the system in this situation.

## New behavior
- Sending a $care-gaps request with no measureId, measureUrl, or measureIdentifier will no longer throw an error, but will run $care-gaps for all measures on the system
- `npm run connectathon-upload` will now ignore `EXM347-4.3.000-bundle.json`, `EXM529-1.0.000-bundle.json`, `EXM349-2.10.000-bundle.json`, `EXM111-9.1.000-bundle.json`, `EXM149-9.2.000-bundle.json`, as they all contain errors that prevent valid gaps calculation
- Can now run $care-gaps operations on measures EXM104, EXM108, and EXM124, all of which have invalid library references, which previously tripped up the test server

## Code changes
- Updated `careGaps` function in measure.service.js to search for all measures on system and run a gaps evaluation for each patient in the subject for each measure
- Removed error throw in `validateCareGapsParams` to allow for no measure identification parameter
- Added bundles `EXM347-4.3.000-bundle.json`, `EXM529-1.0.000-bundle.json`, `EXM349-2.10.000-bundle.json`, `EXM111-9.1.000-bundle.json`, `EXM149-9.2.000-bundle.json` to the ignored list

# Testing guidance
- Run `npm install` to ensure you grab the newly updated fqm-execution dependency
- Run `npm test`
- Run `npm run db:reset`
- Run `npm run connectathon-upload` this will omit the newly ignored bundles
- Run `npm start`
- Submit a `$care-gaps` request without a `measureId`, `measureIdentifier`, or `measureUrl`. For more info on $care-gaps requests, see [here](https://github.com/projecttacoma/deqm-test-server/pull/71)
- Ensure the `parameter` attribute returned is of length 10. most/all of these entries will be empty
- Try testing this with the new [$care-gaps group support](https://github.com/projecttacoma/deqm-test-server/pull/89). You should see an entry in the response parameter array for each patient for each measure (i.e. for group size = 2, and number of measures = 10, parameter length should equal 20)
- Try testing with both `POST` and `GET` requests as well

### DIFF
--- a/src/scripts/getConnectathonBundles.js
+++ b/src/scripts/getConnectathonBundles.js
@@ -4,14 +4,15 @@ const mongoUtil = require('../database/connection');
 const { createResource } = require('../database/dbOperations');
 
 const connectathonPath = path.resolve(path.join(__dirname, '../../connectathon/fhir401/bundles/measure/'));
-// bundles that are not the latest available version
+// bundles that are not the latest available version or contain errors that prevent valid gaps calculation
 const IGNORED_BUNDLES = [
   'EXM347-3.2.000-bundle.json',
   'EXM347-4.3.000-bundle.json',
   'EXM529-1.0.000-bundle.json',
   'EXM349-2.10.000-bundle.json',
   'EXM111-9.1.000-bundle.json',
-  'EXM149-9.2.000-bundle.json'
+  'EXM149-9.2.000-bundle.json',
+  'EXM124-8.2.000-bundle.json'
 ];
 
 // files containing EXM bundles of interest from connectathon repo
--- a/src/scripts/getConnectathonBundles.js
+++ b/src/scripts/getConnectathonBundles.js
@@ -6,6 +6,12 @@ const { createResource } = require('../database/dbOperations');
 const connectathonPath = path.resolve(path.join(__dirname, '../../connectathon/fhir401/bundles/measure/'));
 // bundles that are not the latest available version
 const IGNORED_BUNDLES = ['EXM347-3.2.000-bundle.json'];
+//   'EXM347-4.3.000-bundle.json',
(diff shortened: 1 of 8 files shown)


src/scripts/getConnectathonBundles.js
```diff
@@ -5,7 +5,15 @@ const { createResource } = require('../database/dbOperations');
 
 const connectathonPath = path.resolve(path.join(__dirname, '../../connectathon/fhir401/bundles/measure/'));
 // bundles that are not the latest available version
```
p9g, line 7:
```suggestion
// bundles that are not the latest available version or contain errors that prevent valid gaps calculation
```

sarahmcdougall (changes requested):
Nice work overall with this task, especially considering the hassle of working with the outdated connectathon URLs! 

Aside from some small comments/suggestions that I left, I was wondering if we want to have all these empty objects being returned since they have no context. If someone uses the EXM130 denominator patient as the subject, then only one `parameter` entry will actually have the `resource` populated, so would it make more sense to just return the one populated result? The spec didn't really have any examples for this, so I'm not really sure what the correct approach is. I just feel like with the current output, it can be confusing to know which measures were actually calculated since there is no context in the result. I remember talking with @mgramigna last week-ish about eventually updating fqm-execution so that it no longer returns just an empty object for cases like the EXM130 numerator patient. Maybe we can create a task that will return the measureId for the measure being calculated rather than an empty object?

PR #91, opened 2022-03-03
Add error utility that extends from ServerError

sarahmcdougall:
# Summary
A utility was added to the test server for throwing different errors, which allows us to avoid repeating the `ServerError` object throughout the repository.

## New behavior
Behavior is unchanged.

## Code changes
This new utility is available in `errorUtils.js`. The `CustomServerError` class extends from the `ServerError` class and creates the `options` object that contains the `severity`, `code`, and `details` attributes. Then, various classes extend from the `CustomServerError` class to fill in the details of the `statusCode`, `code`, and `details.text` (the message that we want contained in the error).

Throughout the codebase, the `ServerError` instances have been replaced with the appropriate child error classes. Now, when throwing an error, the only parameter needed (for most cases) is the message that we want contained in the error.

# Testing Guidance
Run all unit tests.
Look through the codebase for instances where each of the different errors are thrown. Try replicating each of these error scenarios in Insomnia to ensure that the error functionality is unchanged.

Here are some examples:
* To throw a BadRequest error, send a GET request to `$evaluate-measure` with an invalid `reportType`
* To throw a NotImplementedError, send a GET request to `$evaluate-measure` with the `subject-list` `reportType`

Check that all the direct `ServerError` calls have been removed from the codebase.

### DIFF
--- a/src/resources/transactionBundle.js
+++ b/src/resources/transactionBundle.js
@@ -3,7 +3,8 @@
  * a txn bundle in the proper format and convert the representation to JSON
  */
 
-const { ServerError, resolveSchema } = require('@projecttacoma/node-fhir-server-core');
+const { resolveSchema } = require('@projecttacoma/node-fhir-server-core');
+const { UnprocessableEntityError } = require('../util/errorUtils');
 
 const createTransactionBundleClass = baseVersion => {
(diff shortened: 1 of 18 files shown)


MegRiley (approved):

MegRiley:
Looks good to me

PR #92, opened 2022-03-07
Gaps Organization Param

natjoe4:
# Summary
The spec allows an organization parameter to be passed in instead of a subject parameter for gaps requests. Passing in this organization parameter should cause a gaps report to be run for each patient which references the organization as their managingOrganization.

## New behavior
The test server now accepts an organization id as a parameter in lieu of a subject for $care-gaps requests. Adding both will cause a `400:BadRequest` error

## Code changes
- Updated validationUtils/validateCareGapsParams to accept an organization param or a subject param
- Created a new util function in measure.service that takes in the query and determines how to find patient ids based on the query params, then searches for those patients in the db
- Updated testing
- Updates to README

# Testing guidance
- Give it an `npm test`
- run `npm run db:reset`, then `npm run connectathon-upload`
- run `npm start`
- `PUT` an new organization onto the server using url: `http://localhost:3000/4_0_1/Organization/1` and body:
```
{
	"resourceType": "Organization",
	"id": "1"
}
```
- Run a `$care-gaps` calculation by `POST`ing this `http://localhost:3000/4_0_1/Measure/$care-gaps?periodStart=2019-01-01&periodEnd=2019-12-31&organization=Organization/1&status=open-gap`
- You should receive an empty parameters object back as no patients reference that organization yet
- Now, make the denominator patient for EXM130 and EXM124 reference our new organization using a `PUT` request to `http://localhost:3000/4_0_1/Patient/denom-EXM130` with body 
```
{
	"resourceType": "Patient",
	"id": "denom-EXM130",
	"managingOrganization": {
		"identifier": {
			"value": "Organization/1"
		}
	}
}
```
- Do the same with denom-EXM124 by replacing the ids in the url and body with `EXM124`
- Now, rerun the gaps command above. You should see two denom patients in the output params now!
- Experiment with param combinations you expect to break the server (subject and an org, org of wrong format, org that doesn't exist, etc...) and ensure the outputs make sense.


lmd59 (commented):
Looks pretty good. A couple of small thoughts!

src/util/patientUtils.js
```diff
@@ -71,4 +71,31 @@ async function getPatientData(patientId, dataRequirements) {
   return data;
 }
 
-module.exports = { getPatientDataCollectionBundle, getPatientDataSearchSetBundle, getPatientData };
+/**
+ * Takes in a care-gaps query and retrieves the ids of the patients to be run against the measure
+ * NOTE: Currently, this function assumes ResourceType/ResourceId-style references
+ * @param {string} subject A reference to either the patient or group resource to run against a measure
+ * @param {string} organization A reference to an organization that
```
lmd59, line 78: Missing end of thought: "A reference to an organization that"...?

lmd59 (approved):

PR #93, opened 2022-03-09
Adding support for practitioner as a gaps in care gaps param

MegRiley:
The spec allows an practitioner  parameter to be passed in instead of a subject parameter for gaps requests. 

New behavior

The test server now accepts an practitioner id as a parameter in lieu of a subject for $care-gaps requests. Adding both will cause a 400:BadRequest error

Code changes

Updated validationUtils/validateCareGapsParams to accept an practitioner param or a subject param but not both
Created a new util function in measure.service that takes in the query and determines how to find patient ids based on the query params, then searches for those patients in the db
Updated testing
Updates to README
Testing guidance

Give it an npm test
run npm run db:reset, then npm run connectathon-upload
run npm start
PUT an new practitioner onto the server using url: http://localhost:3000/4_0_1/Practitioner/1 and body:
{
	"resourceType": "Practitioner",
	"id": "1"
}
Run a $care-gaps calculation by POSTing this http://localhost:3000/4_0_1/Measure/$care-gaps?periodStart=2019-01-01&periodEnd=2019-12-31&practitioner=Practitioner/1&status=open-gap
You should receive an empty parameters object back as no patients reference that practitioner yet
Now, make the denominator patient for EXM130 and EXM124 reference our new practitioner using a PUT request to http://localhost:3000/4_0_1/Patient/denom-EXM130 with body
{
	"resourceType": "Patient",
	"id": "denom-EXM130",
	"generalPractitioner": {
		"identifier": {
			"value": "Practitioner/1"
		}
	}
}
Do the same with denom-EXM124 by replacing the ids in the url and body with EXM124
Now, rerun the gaps command above. You should see two denom patients in the output params now!


natjoe4 (changes requested):
A couple things to look at here. I think the description of how practitioner was supposed to be handled was a little off in the task description. Cleared things up with Matt, but there are a few places in the code that don't quite process as intended.

src/util/patientUtils.js
```diff
@@ -93,6 +93,9 @@ const retrievePatientIds = async ({ subject, organization }) => {
     return referencedObject.member.map(m => m.entity.reference.split('/')[1]);
   } else if (reference[0] === 'Patient') {
     return [subject.split('/')[1]];
+  } else if (reference[0] === 'Practitioner') {
```
natjoe4, line 96: I think you're missing some logic here. Reference will never equal practitioner since the split statement on 84 can only be subject or organization. Also, since practitioner can only be provided with an organization, this conditional should be inside the organization conditional, and the query should check that the patients generalPractitioner is equal to practitioner and that the managingOrganization is equal to organization

src/util/validationUtils.js
```diff
@@ -78,12 +78,22 @@ const validateCareGapsParams = query => {
...
+    if (query.practitioner) {
+      throw new BadRequestError('Must provide either subject or practitioner. Received both');
+    }
     const subjectReference = query.subject.split('/');
     if (subjectReference.length !== 2 || !['Group', 'Patient'].includes(subjectReference[0])) {
       throw new BadRequestError(
         `Subject may only be a Group resource of format "Group/{id}" or Patient resource of format "Patient/{id}".`
       );
     }
+  } else if (query.practitioner) {
```
natjoe4, line 90: This case should actually throw an error, since we can only reach here if a practitioner was provided without and organization. I would move the contents of this conditional inside the conditional that checks for an organization being passed, and just have this conditional throw an error saying practitioner must be passed in with organization

test/util/validationUtils.test.js
```diff
@@ -279,6 +302,35 @@ describe('validateCareGapsParams', () => {
...
+  test('validateCareGapsParams does not throw error with practitioner instead of subject', async () => {
+    const VALID_REQ = {
+      query: VALID_PRACTITIONER_QUERY
+    };
+    expect(validateCareGapsParams(VALID_REQ.query)).toBeUndefined();
+  });
+  test('validateCareGapsParams throws error with both practitioner and subject', async () => {
+    try {
+      validateCareGapsParams(SUBJECT_AND_PRACTITIONER_QUERY);
+      throw new Error('validateCareGapsParams failed to throw an error when provided both subject and practitioner');
```
natjoe4, line 314: Also my bad, but instead of throwing an error here, use expect.fail() with the same message. Will make debugging easier

src/util/validationUtils.js
```diff
@@ -78,12 +78,22 @@ const validateCareGapsParams = query => {
       );
     }
   } else if (query.subject) {
+    if (query.practitioner) {
+      throw new BadRequestError('Must provide either subject or practitioner. Received both');
```
natjoe4, line 82: I'd rephrase this error message. subject and practitioner are not direct replacements for each other. Maybe try something simple like "Cannot provide both a subject and practitioner"

natjoe4 (changes requested):
A couple things to look at here. I think the description of how practitioner was supposed to be handled was a little off in the task description. Cleared things up with Matt, but there are a few places in the code that don't quite process as intended.

test/util/validationUtils.test.js
```diff
@@ -279,6 +310,38 @@ describe('validateCareGapsParams', () => {
...
+      expect(e.issue[0].details.text).toEqual('Cannot provide both a subject and practitioner');
+    }
+  });
+  test('validateCareGapsParams throws error with invalid practitioner format', async () => {
+    try {
+      expect(validateCareGapsParams(INVALID_PRACTITIONER_QUERY)).toBeUndefined();
+    } catch (e) {
+      expect(e.statusCode).toEqual(400);
+      expect(e.issue[0].details.text).toEqual(
+        'Practitioner may only be an Practitioner resource of format "Practitioner/{id}". Received: INVALID'
```
natjoe4, line 340:
```suggestion
        'Practitioner may only be a Practitioner resource of format "Practitioner/{id}". Received: INVALID'
```

test/util/validationUtils.test.js
```diff
@@ -39,7 +39,38 @@ const SUBJECT_AND_ORGANIZATION_QUERY = {
...
-
+const SUBJECT_AND_PRACTITIONER_QUERY = {
+  periodStart: '2019-01-01',
+  periodEnd: '2019-12-31',
+  status: 'open-gap',
+  practitioner: 'Practitioner/testPractitioner',
+  subject: 'Patient/testPatient',
+  measureId: 'testID'
+};
+const PRACTITIONER_AND_NO_ORG = {
```
natjoe4, line 50: Sorry didn't catch this before, but isn't this fixture the same as the one above? I'd remove the subject property here and update the test to make sure you're getting the error case you want.

natjoe4 (changes requested):
Just a couple more things on the tests just added. Matt also mentioned this morning that the commit logs include my commits from the branch this was initially based on. Can you do an interactive rebase with main that drops all my commits so that we don't get repeats in the main Gitlog? Feel free to reach out if you'd like to go over that together.

natjoe4 (approved):
Looks good to me!

PR #94, opened 2022-03-09
Dependent library missing descriptive error

natjoe4:
# Summary
Adds a descriptive server error throw when a dependent library for an $evaluate-measure request cannot be found

## New behavior
Previously, when a dependent library could not be found, an error would be thrown with message: `"Unexpected: Cannot read property 'id' of null"`. The current message now details the id or canonical url of the missing library as well as the version when appropriate

## Code changes
- Updated `getAllDependentLibraries()` function to check whether the found lib is null and throw an error if so
- Added testing

# Testing guidance
- Run `npm test`
- Run `npm run db:reset`
- Run `npm run connectathon-upload`
- Run `npm start`
- Send a `DELETE` request to `http://localhost:3000/4_0_1/Library/library-FHIRHelpers-4.0.1` and ensure `204` is returned
- Send a `POST` request to `http://localhost:3000/4_0_1/Measure/measure-EXM130-7.3.000/$evaluate-measure?subject=numer-EXM130&periodStart=2019-01-01&periodEnd=2019-12-31` and ensure that a `500` response code is returned with an appropriate error message.

### DIFF
--- a/src/util/bundleUtils.js
+++ b/src/util/bundleUtils.js
@@ -192,7 +192,9 @@ async function getAllDependentLibraries(lib) {
     const lib = await findOneResourceWithQuery(libQuery, 'Library');
     if (lib === null) {
       throw new InternalError(
-        `Failed to find dependent library with canonical url: ${libQuery.url} and version: ${libQuery.version}`
+        `Failed to find dependent library with ${
+          libQuery.id ? `id: ${libQuery.id}` : `canonical url: ${libQuery.url}`
+        }${libQuery.version ? `and version: ${libQuery.version}` : ''}`
       );
     }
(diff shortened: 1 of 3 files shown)


sarahmcdougall (changes requested):
Super helpful having this error thrown now. I like the inclusion of logic for the id/canonical URL/version - it allows us to be as specific as possible.

The unit test failed for me because the `testLibraryDependencies.json` file could not be found for the test - I think it may have gotten lost in the commits because I don't see it in the repo either. I think that is what's accounting for the GH Actions still running. I can take another look once that file is added!

sarahmcdougall (approved):

PR #97, opened 2022-03-28
Validation Env

natjoe4:
# Summary
DEQM test server now supports optional FHIR validation through Inferno's validation server!

## New behavior
When the VALIDATE, VALIDATION_HOST, and VALIDATION_URL are all set appropriately, DEQM test server will now query the Inferno FHIR validation server to ensure the bodies of all POST and PUT requests which attempt to add to or alter data on the server are valid FHIR. Also updates the meta.profile of the submitted resource to contain all profiles validated against

## Code changes
- Added validation middleware
- Added resourceValidationUtils file
- validationUtils.js -> operationValidationUtils.js
- Changed server initialize function to async
- Updated tests to use new async initialize function
- Added testing for resourceValidationUtils

# Testing guidance
- `npm run test`
- run the Inferno FHIR validator with `docker run -p 4567:4567 infernocommunity/fhir-validator-service`
- In a new terminal, start up the test server using `VALIDATE=true VALIDATION_HOST=localhost VALIDATION_PORT=4567 npm run start`
- Post/Put resources to the server, both valid and invalid. Ensure that the errors you get are intuitive and the meta.profile of the submitted resource is updated to include the base resource type (i.e. Patient, Observation... etc) as well as the qicore resource type when the resource is valid against qicore
- Try using resources from connectathon bundles, especially those with included meta.profiles and ensure that the validator validates against those profiles as well. NOTE: most of these resources (if not all) are actually invalid and will fail.
- If you have issues finding resources to try, just ping me! Happy to show you what I used.
- Run measure operations as well as GET and DELETE requests and ensure they are unaffected.


src/server/server.js
```diff
@@ -22,10 +23,21 @@ class DEQMServer extends Server {
     this.app.get('/:base_version/file/:clientId/:fileName', configClientFile.clientFile);
     return this;
   }
+  async enableValidationMiddleWare() {
+    await axios.put('http://localhost:4567/igs/hl7.fhir.us.qicore');
+    this.app.put('/:base_version*', validateFhir);
+    this.app.post('/:base_version*', validateFhir);
```
natjoe4, line 29: Can we be sure the async logic will work correctly here? I don't totally understand how the workflow of async middleware works

mgramigna: This `PUT` is just loading QI core into the validator, right? Do we _need_ to do it in the declaration of this middleware? Can we not actually just do it in `index.js` when the server spins up? We do `await mongoClient.connect()` as an async thing after the server starts, can we just do it after that? 

Also, should use `process.env.VALIDATOR_HOST/PORT` heheh. I guess I'll just review this PR

src/util/resourceValidationUtils.js
```diff
@@ -0,0 +1,124 @@
...
+        code: 'exception',
+        details: {
+          text: `Validation server failed while validating against profiles: ${profiles.join(',')}`
+        }
+      }
+    ];
+    const OperationOutcome = resolveSchema(base_version, 'operationoutcome');
+    return { isValid: false, code: 500, data: new OperationOutcome(outcome).toJSON() };
+  }
+  if (response.data.issue[0].details.text !== 'All OK') {
```
mgramigna, line 118: I think we should be a bit more generous here with what classifies validation being fine. There are cases where the validator will issue "informational" or "warning" level "issues", which I don't think will classify as "Bad Request". I think in later tasking we can think about how if at all we want to send those warnings back, but for now I think we should let them through.

A way to replicate this behavior is to change the URL of a code system on some coding (e.g. I changed a Patient resource's `identifier[0].type.coding[0].system` to be `http://example.com`. You'll see that we don't get "All Ok", but we also don't get anything with severity "error".

All this is to say I think we should make a helper that takes in the validation response and returns a bool for if it should pass or not. I think right now it should return true if the issue array contains _no_ objects with severity `error`, and false if any are present.

Let's unit test this baby in isolation too. It might change later

src/util/resourceValidationUtils.js
```diff
@@ -0,0 +1,124 @@
+const { resolveSchema } = require('@projecttacoma/node-fhir-server-core');
```
mgramigna, line 1: Let's add some logger debug bois to this file, maybe some infos as well. Would be nice to know what's happening along the way

src/util/resourceValidationUtils.js
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Checks body of request by querying a FHIR validation server. Updates the submitted resource to
+ * include validated profiles in the meta.profile field. Responds with 400 if body is not valid FHIR
+ * 500 if validation server error
+ * @param {Object} req request object of an express request
+ * @param {Object} res response object of an express request
+ * @param {Function} next express function which passes info along after clearing the middleware
+ */
+async function validateFhir(req, res, next) {
```
mgramigna, line 13: A lot of this function is uncovered by unit tests. Any way we can try to bump the numbers up? Hopefully we can just mimick the inputs easily. If it's too much then no worries.

src/util/resourceValidationUtils.js
```diff
@@ -0,0 +1,124 @@
...
+  const metaProfiles = body?.meta?.profile;
+  if (metaProfiles) {
+    profiles.push(...metaProfiles);
+  }
+  if (params[params.length - 1] === '$submit-data') {
+    profiles.push('Parameters');
+  }
+  // We don't need to validate posted Parameters bodies for dollar-sign operations since these aren't stored in the db
+  else if (params[params.length - 1][0] === '$') {
+    return 'Calculation';
```
mgramigna, line 70: I see that we use this `'Calculation'` string to indicate when we don't care about doing validation (the whole validation logic above is wrapped in an `if (profiles !== 'Calculation')`. This confused me a bit at first and I had to trace it down to this function.

I think it would be a little better to just return `[]` here, as that makes more conceptual sense for this function: you ask it to retrieve the profiles you want to validate against, and it certain cases you give it nothing back. Then, the above if statement can just be `if (profiles.length > 0)`

src/util/resourceValidationUtils.js
```diff
@@ -0,0 +1,124 @@
...
+    process.env.VALIDATOR_PORT
+  }/validate?profile=${profiles.join(',')}`;
+  let response;
+  try {
+    response = await axios.post(validationUrl, body);
+  } catch (e) {
+    /*
+        We can't throw server errors inside this middleware, so best to construct
+        our own operation outcomes and return them
+    */
```
mgramigna, line 103: rip

test/util/resourceValidationUtils.test.js
```diff
@@ -0,0 +1,59 @@
+const { retrieveProfiles, getValidationInfo } = require('../../src/util/resourceValidationUtils');
+
+describe('resourceValidationInfo tests', () => {
+  test('retrieveProfiles returns Calculation on dollar-sign operation', () => {
+    const profile = retrieveProfiles('/4_0_1/Measure/$dataRequirements', {});
+    expect(profile).toEqual('Calculation');
```
mgramigna, line 6: this will need to update based on my above comment

test/util/resourceValidationUtils.test.js
```diff
@@ -0,0 +1,59 @@
+const { retrieveProfiles, getValidationInfo } = require('../../src/util/resourceValidationUtils');
+
+describe('resourceValidationInfo tests', () => {
+  test('retrieveProfiles returns Calculation on dollar-sign operation', () => {
+    const profile = retrieveProfiles('/4_0_1/Measure/$dataRequirements', {});
```
mgramigna, line 5: nitpick lol
```suggestion
    const profile = retrieveProfiles('/4_0_1/Measure/$data-requirements', {});
```

test/util/resourceValidationUtils.test.js
```diff
@@ -0,0 +1,59 @@
...
+  test('retrieveProfiles returns Patient with Patient URL', () => {
+    const profile = retrieveProfiles('/4_0_1/Patient', {});
+    expect(profile).toEqual(['Patient'], {});
+  });
+  test('retrieveProfiles returns Patient and meta.profile info with Patient URL and meta.profile in body', () => {
+    const profile = retrieveProfiles('/4_0_1/Patient', { meta: { profile: ['testProfile', 'testProfile2'] } });
+    expect(profile).toEqual(['testProfile', 'testProfile2', 'Patient'], {});
+  });
+  test('getValidationInfo returns correct object on valid FHIR response', async () => {
+    const axios = require('axios');
```
mgramigna, line 21: Do we need to require axios twice (here and below) in order for the mocks to works expected? If so I think there's a better solution we can do where we only require axios once at the top of the file, and then do an `afterEach` hook to clear all configured mocks in between tests

mgramigna (changes requested):
This is sick. Very sleek, performant, and well written. I have a bunch of inline comments, but great work overall

src/util/resourceValidationUtils.js
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Checks body of request by querying a FHIR validation server. Updates the submitted resource to
+ * include validated profiles in the meta.profile field. Responds with 400 if body is not valid FHIR
+ * 500 if validation server error
+ * @param {Object} req request object of an express request
+ * @param {Object} res response object of an express request
+ * @param {Function} next express function which passes info along after clearing the middleware
+ */
+async function validateFhir(req, res, next) {
```
mgramigna, line 14: Nitpick, but if request type is not `PUT` or `POST`, we should be able to return `next()` early. Not crucial, but saves us a tiny morsel of computation

mgramigna (changes requested):

src/util/resourceValidationUtils.js
mgramigna: actually, this comment might be irrelevant since this call should never happen if the request type is anything except PUT or POST, right? Since that's how the middleware is configured. Yeah, nvm I think you can ignore this!

PR #98, opened 2022-03-29
Initial changes

MegRiley:
(unit tests are still broken)

# Summary

## Code changes

# Testing guidance


test/util/validationUtils.test.js
```diff
@@ -388,20 +388,15 @@ describe('validateCareGapsParams', () => {
...
       validateCareGapsParams(INVALID_REQ.query);
-      throw new Error(
-        'validateCareGapsParams failed to throw an error when provided both program and measure identification'
-      );
     } catch (e) {
+      console.log(JSON.stringify(e));
       expect(e.statusCode).toEqual(501);
-      expect(e.issue[0].details.text).toEqual(
-        'Simultaneous program and measure identification (measureId/measureIdentifier/measureUrl) is not currently supported by the server.'
-      );
```
natjoe4, line 404: This test isn't quite set up right. Right now it will pass if validateCare Gaps throws no error or a 501. What we want is to wrap the validateCareGaps call in an expect().toBeUndefined. We can actually get rid of the whole try/catch block, since if an error is thrown, the test will fail with a comprehensive message anyway.

natjoe4, line 404: This test isn't quite set up right. Right now it will pass if validateCare Gaps throws no error or a 501. What we want is to wrap the validateCareGaps call in an expect().toBeUndefined. We can actually get rid of the whole try/catch block, since if an error is thrown, the test will fail with a comprehensive message anyway.

natjoe4 (changes requested):
Mainly smaller changes. Also, Calculator.calculateDataRequirements recently became async, so this will need to be awaited in care-gaps.

test/util/operationValidationUtils.test.js
```diff
@@ -388,21 +388,11 @@ describe('validateCareGapsParams', () => {
     }
   });
 
-  test('validateCareGapsParams throws error with both program and measure identification', async () => {
+  test('validateCareGapsParams does not throw an error with both program and measure identification', async () => {
     const INVALID_REQ = {
```
natjoe4, line 392: Super picky here, but this req is no longer invalid, so should probably be renamed

src/services/measure.service.js
```diff
@@ -392,8 +411,9 @@ const careGaps = async (args, { req }) => {
         }
       })
     };
+
     // TODO: add any searchTerm (measure identifier) query ANDed with above query
```
natjoe4, line 415:
```suggestion
```

src/services/measure.service.js
```diff
@@ -383,6 +383,25 @@ const careGaps = async (args, { req }) => {
...
     const progArr = Array.isArray(query.program) ? query.program : [query.program];
     // build query AND-ing all potential program parameters
+
+    let measureQuery = {};
+    if (searchTerm) {
+      let prop = Object.keys(searchTerm)[0];
+
+      // for now assume we only support one of a possible identifier property
+
+      if (Array.isArray(searchTerm[prop])) {
```
natjoe4, line 393: This only covers the situation where the user passes in an array of measure identifiers though. I think we still need to use the search term even if it is not an array

natjoe4 (changes requested):
All functional and looks good except one small edge case. Also note the lint error.

src/services/measure.service.js
```diff
@@ -513,10 +529,13 @@ const systemCodeProgramQuery = program => {
  * @param {Object} query http request query
```
natjoe4, line 529: this jsdoc needs an update to account for the added flag. Also, can we switch the flag to camelCase? isForQB

src/services/measure.service.js
```diff
@@ -513,10 +529,13 @@ const systemCodeProgramQuery = program => {
...
  * @returns {Object} an object containing the measure identifier with the appropriate key
  */
-const retrieveSearchTerm = query => {
+const retrieveSearchTerm = (query, isforqb) => {
   const { measureId, measureIdentifier, measureUrl } = query;
   if (measureId) {
-    return { _id: measureId };
+    //some manipulation will be needed here because _id means a generated id when interacting with mongo
+    //however if this field is used with the asymetrik query builder it means the actual id of the measure
```
natjoe4, line 536:
```suggestion
    //however if this field is used with the Asymmetrik query builder it means the actual id of the measure
```

natjoe4 (changes requested):
Two tiny changes and then it should be ready for approval!

PR #102, opened 2022-04-08
Clean up test fix setup to accept an array 

MegRiley:
# Summary
Changed `testSetup` to no longer accept a fixed list of 3 items but to now accept an array of items that can be updated for new tests as resources are added
## New behavior
The behavior should be identical to prior behavior however it should make the beforeAll and beforeEach blocks cleaner as all that is needed to add a new resource for a test 
## Code changes
 `test/populateTestData.js` - modified `testSetup` to now accept an array of resources and add them to the database
`test/services/base.service.test.js` - updated to use the new `testSetup` call and remove the unneeded method
`test/services/measure.service.test.js`   -updated to use the new `testSetup` call and remove the unneeded method
`test/services/patient.service.test.js` -updated to use the new `testSetup` call and remove the unneeded method

# Testing guidance
 Run npm test and assure all the tests pass. Ultimately the behavior should be the same.

### DIFF
--- a/test/services/measure.service.test.js
+++ b/test/services/measure.service.test.js
@@ -16,7 +16,7 @@ const testParamInvalidResourceType = require('../fixtures/fhir-resources/paramet
 const testEmptyParam = require('../fixtures/fhir-resources/parameters/emptyParam.json');
 const testParamTwoMeasureReports = require('../fixtures/fhir-resources/parameters/paramTwoMeasureReports.json');
 const testCareGapsMeasureReport = require('../fixtures/testCareGapsMeasureReport.json');
-const { testSetup, cleanUpTest, createTestResource } = require('../populateTestData');
+const { testSetup, cleanUpTest } = require('../populateTestData');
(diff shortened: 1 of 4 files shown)


test/services/measure.service.test.js
```diff
@@ -30,13 +29,20 @@ describe('measure.service', () => {
...
+      testMeasure,
+      testMeasure2,
+      testPatient,
+      testPatient2,
+      testLibrary,
+      testGroup,
+      testOrganization,
+      testOrganization2,
+      testParam,
+      testParamInvalidResourceType
```
hossenlopp, line 42: `testParam`, `testParamInvalidResourceType` are not resources that should be added to the database as they are used as parameters for tests. 

`deleteMeasure` should be added to this list since it is a measure that is intended to be deleted. Which actually the test that uses it should be updated to ensure the Measure Resource was removed.

hossenlopp (changes requested):

hossenlopp (approved):

PR #106, opened 2022-04-28
Updated Base Fhir Validation profile

natjoe4:
# Summary
Previously, the validation layer would append the base fhir resourceType to the meta.profile of a validated resource. Now, it append the full StructureDefinition url `http://hl7.org/fhir/StructureDefinition/<ResourceType>`

## New behavior
When a resource passes validation, the full StructureDefinition URL will be added to its meta.profile instead of the base resource type string

## Code changes
prepended `http://hl7.org/fhir/StructureDefinition/` before base resource type when checking validation

# Testing guidance
- run npm test
- Follow instructions here #97 but ensure that the meta.profile of validated resources contains `http://hl7.org/fhir/StructureDefinition/<ResourceType>` instead of the base resourceType string


sarahmcdougall (approved):
thank you for making these ground-breaking changes :trophy:

PR #108, opened 2022-05-05
May 2022 connectathon changes

sarahmcdougall:
# Summary
Various changes to enable type filter specification in bulk submit data requests.

## New behavior
Now, the user can add a `useTypeFilters` parameter to a bulk submit data request. This is used in bulk data utilities to get the type filters from the data requirements. 

Additional logic was added to detect when a type filter parameter is included in the bulk submit data request - now, the user can add the type filter parameter to the `parameter` array of the bulk submit data request rather than only having the option to specify it in the export url.

## Code changes
* Adds `useTypeFilters` boolean to import worker jobs (defaults to false if not provided)
* Adds search for `useTypeFilters` boolean to `bulkImportFromRequirements()` so that it can be added to the job data if present in the bulk submit data request
* Changes logic in `exportUtils` for parsing through the export url in conjunction with the parameter objects in the `parameter` array of the bulk submit data request. Now goes through the `parameter` array and also filters out the type filters parameter. When appending the parameters to the export Url, first checks if the url already contains the type and/or type filter parameters - if so, throws a console.warn about the parameter already being present and does not use the entries from the `parameter` array.

# Testing guidance
Run the unit tests for working with the type filter parameters.

For testing, you will need to npm link to the `may22-connectathon` branch in bulk data utilities. 
* Start up the test server
* POST a bulk submit data request that contains an exportUrl parameter
* POST a bulk submit data request that contains an exportUrl parameter and a `useTypeFilters` parameter. You should notice that less resources are imported

```json
{
     "name": "useTypeFilters", 
     "valueBoolean": true
}
```
* POST some bulk data requests with and without type and type filter parameters, and with these parameters included in the actual export url

### DIFF
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -181,10 +181,19 @@ const bulkImportFromRequirements = async (args, { req }) => {
 
   // retrieve data requirements
   const exportURL = retrieveExportUrl(parameters);
+
+  // retrieve useTypeFilters boolean
+  const useTypeFiltersArray = parameters.filter(param => param.name === 'useTypeFilters');
+  let useTypeFilters;
+  if (useTypeFiltersArray.length > 0) {
+    useTypeFilters = useTypeFiltersArray[0].valueBoolean;
+  }
+
   const jobData = {
     clientEntry,
     exportURL,
-    measureBundle
+    measureBundle,
+    useTypeFilters
   };
   await importQueue.createJob(jobData).save();
   res.status(202);
--- a/src/util/exportUtils.js
+++ b/src/util/exportUtils.js
@@ -22,10 +22,31 @@ const retrieveExportUrl = parameters => {
     })
     .toString();
 
+  const typeFilterString = parameters
+    .filter(param => param.name === '_typeFilter')
+    .map(function (typeFilter) {
+      logger.debug(`Adding typeFilter ${typeFilter} to exportUrl typeFilter parameter`);
+      return typeFilter.valueString;
+    })
+    .toString();
+
   if (typesString) {
-    exportUrl = `${exportUrl}?_type=${typesString}`;
+    if (exportUrl.includes(`_type=`)) {
+      console.warn('_type parameter already supplied in exportUrl. Omitting entries from parameter array');
+    } else {
+      // add types from parameters to exportUrl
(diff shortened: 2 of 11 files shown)


src/services/measure.service.js
```diff
@@ -181,10 +181,19 @@ const bulkImportFromRequirements = async (args, { req }) => {
 
   // retrieve data requirements
   const exportURL = retrieveExportUrl(parameters);
+
+  // retrieve useTypeFilters boolean
+  const useTypeFiltersArray = parameters.filter(param => param.name === 'useTypeFilters');
+  let useTypeFilters;
+  if (useTypeFiltersArray.length > 0) {
+    useTypeFilters = useTypeFiltersArray[0].valueBoolean;
```
hossenlopp, line 189: This line of code does not appear to be covered by unit tests.

src/util/exportUtils.js
```diff
@@ -22,10 +22,31 @@ const retrieveExportUrl = parameters => {
...
+    .map(function (typeFilter) {
+      logger.debug(`Adding typeFilter ${typeFilter} to exportUrl typeFilter parameter`);
+      return typeFilter.valueString;
+    })
+    .toString();
+
   if (typesString) {
-    exportUrl = `${exportUrl}?_type=${typesString}`;
+    if (exportUrl.includes(`_type=`)) {
+      console.warn('_type parameter already supplied in exportUrl. Omitting entries from parameter array');
```
hossenlopp, line 35: This situation isn't tested by unit tests.

src/util/exportUtils.js
```diff
@@ -22,10 +22,31 @@ const retrieveExportUrl = parameters => {
...
+    } else {
+      // add types from parameters to exportUrl
+      exportUrl += `${exportUrl.includes('_typeFilter=') ? '&' : '?'}_type=${typesString}`;
+    }
   }
 
+  if (typeFilterString) {
+    if (exportUrl.includes(`_typeFilter=`)) {
+      console.warn('_typeFilter parameter already supplied in exportUrl. Omitting entries from parameter array');
```
hossenlopp, line 44: This situation isn't tested by unit tests.

hossenlopp (commented):
Update the `bulk-data-utilites` and this will be ready for merge. If possible add a test to boost coverage a bit.

hossenlopp (approved):
Hope the CI failure is just a quirk of that GH action.

PR #109, opened 2022-06-09
Added meta.lastUpdated to created and updated resources

sakhin123:
# Summary
- Added meta.lastUpdated tag as FHIR instant to resources for update and create requests
## New behavior
- When updating or creating a resource, meta.lastUpdated tag gets filled/updated to be current time in GMT
## Code changes
- getCurrentInstant() method added to src/util/baseUtils , which returns current date/time as string
- baseCreate() and baseUpdate() methods in src/service/base.service changed to include meta.lastUpdated set to getCurrentInstant()
- added testing to ensure meta.lastUpdated is added to update and create requests
# Testing guidance
- run `npm run test`
- Send post or put request creating or updating a resource. DO NOT include a meta.lastUpdated field
- Send GET request to the created or updated resource and ensure that the meta.lastUpdated reflects when the resource was created or last updated

### DIFF
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -11,7 +11,7 @@ const {
   findResourcesWithAggregation
 } = require('../database/dbOperations');
 const { checkProvenanceHeader, populateProvenanceTarget } = require('../util/provenanceUtils');
-const { checkSupportedResource, checkContentTypeHeader } = require('../util/baseUtils');
+const { checkSupportedResource, checkContentTypeHeader, getCurrentInstant} = require('../util/baseUtils');
 const logger = require('../server/logger.js');
 
 /**
@@ -91,6 +91,7 @@ const baseCreate = async ({ req }, resourceType) => {
   checkSupportedResource(data.resourceType);
   //Create a new id regardless of whether one is passed
   data['id'] = uuidv4();
+  data['meta'] = {lastUpdated: getCurrentInstant(), ...data['meta']};
   if (req.headers['x-provenance']) {
     checkProvenanceHeader(req.headers);
     const res = req.res;
@@ -215,6 +216,7 @@ const baseUpdate = async (args, { req }, resourceType) => {
   if (data.id !== args.id) {
     throw new BadRequestError('Argument id must match request body id for PUT request');
   }
+  data['meta'] = {lastUpdated: getCurrentInstant(), ...data['meta']};
   if (req.headers['x-provenance']) {
     checkProvenanceHeader(req.headers);
     const res = req.res;
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -11,7 +11,7 @@ const {
   findResourcesWithAggregation
 } = require('../database/dbOperations');
 const { checkProvenanceHeader, populateProvenanceTarget } = require('../util/provenanceUtils');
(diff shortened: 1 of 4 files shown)


src/services/base.service.js
```diff
@@ -91,6 +91,7 @@ const baseCreate = async ({ req }, resourceType) => {
   checkSupportedResource(data.resourceType);
   //Create a new id regardless of whether one is passed
   data['id'] = uuidv4();
+  data['meta'] = { lastUpdated: getCurrentInstant(), ...data['meta'] };
```
mgramigna, line 94: Unfortunately this causes a bug where, if `data.meta.lastUpdated` is provided in the request body, the `getCurrentInstant` value gets overwritten since the `...data['meta']` appears later in this object creation.

I'd switch the order of these two object properties and ensure that it overwrites any value that might be provided in the request body and add a comment explaining why the order matters.

I'd also recommend adding a unit test case for this overwriting functionality.

src/services/base.service.js
```diff
@@ -215,6 +216,7 @@ const baseUpdate = async (args, { req }, resourceType) => {
   if (data.id !== args.id) {
     throw new BadRequestError('Argument id must match request body id for PUT request');
   }
+  data['meta'] = { lastUpdated: getCurrentInstant(), ...data['meta'] };
```
mgramigna, line 219: See above comment about overwriting

test/services/base.service.test.js
```diff
@@ -147,6 +167,23 @@ describe('base.service', () => {
...
+        .send(UPDATE_PATIENT_2)
+        .set('Accept', 'application/json+fhir')
+        .set('content-type', 'application/json+fhir')
+        .set('x-provenance', JSON.stringify(SINGLE_AGENT_PROVENANCE))
+        .expect(200)
+        .then(async () => {
+          const patientCollection = db.collection('Patient');
+          const retrievedPatient = await patientCollection.findOne({ id: UPDATE_PATIENT_2.id });
+          expect(retrievedPatient.meta.lastUpdated).toBeDefined();
+          expect(new Date(retrievedPatient.meta.lastUpdated) > new Date(testPatient2.meta.lastUpdated));
```
mgramigna, line 183: Sadly this assertion will always pass regardless of the value inside the parens here, since the jest expression `expect(true)` or `expect(false)` will never throw an error since there isn't a right side of the assertion :/ confusing I know.

We should add a right side here to make sure that the statement inside the `expect` evaluates to true.

```suggestion
          expect(new Date(retrievedPatient.meta.lastUpdated) > new Date(testPatient2.meta.lastUpdated)).toBe(true);
```

mgramigna (changes requested):
Looks great! I found a few nuances that need exploring, but overall the functionality is nice and congrats on your first PR :) 

* Made some inline comments mainly about the assertions and overwriting functionality that need some looking at
* I also noticed that we differ from HAPI in the sense that HAPI doesn't update `meta.lastUpdated` during a PUT request unless the content of the request body has actually changed. This is not something you should address in this PR, but I'm noting it here to think about as a future task.

src/services/base.service.js
sakhin123: Fixed line 94 and 219, added comments about fix

mgramigna (approved):
Woohoo!

PR #111, opened 2022-06-27
Add integration for cql-exec-fhir-mongo

mgramigna:
# Summary

Adds config/functionality for uses of [cql-exec-fhir-mongo](https://github.com/projecttacoma/cql-exec-fhir-mongo) in the test server. Depends on changes from https://github.com/projecttacoma/cql-exec-fhir-mongo/pull/4

## New behavior

New option for scaled calculation to use mongo-based patient source

## Code changes

* Add new option in `.env` and documentation for it
* Add cql-exec-fhir-mongo as a dep (NOTE: will remove branch name when dependent PR is merged)
* Added function in execWorker that resolves the patient data to pass to fqm-execution based on above environment variable

# Testing guidance

Suggested steps to showcase functionality

1) Set up a clean DB with `npm run db:reset` and `npm run connectathon-upload`
2) Reduce the `SCALED_EXEC_THRESHOLD` environment variable to `1` to force scaled calculation
3) Run `$evaluate-measure` on EXM130 or something with reportType population (you should see scaled exec happening in the console) -- note the resulting measure report
4) Switch `.env` to use `mongo` as the `SCALED_EXEC_STRATEGY` and restart server
5) Run the exact same eval measure request in 3) without changing anything -- should get different logs in console, but should get the exact same measure report


sarahmcdougall (approved):
Tested with both the `mongo` and `bundle` strategies and got the same output for `$evaluate-measure`, and the unit tests all pass.

No suggested changes :) We could possibly add a `console.warn` if the user provides an invalid exec strategy (since it will end up defaulting to `bundle` in the exec worker), but I imagine this wouldn't happen that often.

I'm going to approve but hold off on merging until we update the commit hash for the `cql-exec-fhir-mongo` dependency.

PR #113, opened 2022-07-20
Add check to make sure resourceType matches expected type

hossenlopp:
# Summary
Checks that the resourceType of the provided resource matches the expected resourceType for the given resource service/url on create and update calls.

## New behavior
Now returns 400 BadRequest if the resourceType in the body does not match the resourceType of the url.

## Code changes
Added common comparison function that throws an error if the resourceType does not match the expected resourceType. Use that function in baseCreate and baseUpdate functions.

# Testing guidance
- Run unit tests.
- Attempt to POST a mismatch resource. ex. A Patient resource as the body of a request to `/4_0_1/Library` 
  - Response should be 400 and indicate the wrong resource type was found in the body.
- Attempt to PUT a mismatch resource. ex. A Observation resource as the body of a request to `/4_0_1/Patient` 
  - Response should be 400 and indicate the wrong resource type was found in the body.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -117,9 +117,9 @@
           }
         },
         "sanitize-html": {
-          "version": "2.7.0",
-          "resolved": "https://registry.npmjs.org/sanitize-html/-/sanitize-html-2.7.0.tgz",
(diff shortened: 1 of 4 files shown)


sarahmcdougall (approved):
Works great! Glad we throw an error for this now :)

PR #115, opened 2022-07-22
Search Pagination Links

hossenlopp:
# Summary
Pagination of search results was already working. This just adds the links to the searchset bundle so they can easily be used.

## New behavior
`link` attribute in searchset bundle resources is now populated with the `self`, `first`, `previous`, `next`, `last` links as appropiate. For example, on page 3 of 11:

```json
{
  "link": [
    {
      "relation": "self",
      "url": "http://localhost:3000/4_0_1/ValueSet?page=3"
    },
    {
      "relation": "first",
      "url": "http://localhost:3000/4_0_1/ValueSet?page=1"
    },
    {
      "relation": "previous",
      "url": "http://localhost:3000/4_0_1/ValueSet?page=2"
    },
    {
      "relation": "next",
      "url": "http://localhost:3000/4_0_1/ValueSet?page=4"
    },
    {
      "relation": "last",
      "url": "http://localhost:3000/4_0_1/ValueSet?page=11"
    }
  ]
}
```

## Code changes
Added link creation to the search result bundle creation using the existing data from returned results from mongo queries.

# Testing guidance
- Load database with more than 10 of any resource type. Connectahon load should do this for ValueSets.
- Run search on the large resource type. Make sure all appropriate links are in results. 
  - Page sizes are 10 resources and 1 indexed.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+coverage
--- a/src/services/base.service.js
+++ b/src/services/base.service.js
@@ -171,7 +171,6 @@ const baseSearch = async (args, { req }, resourceType, paramDefs) => {
   // build the aggregation query
   logger.debug('Building search query');
   const filter = qb.buildSearchQuery({ req: req, includeArchived: true, parameterDefinitions: searchParams });
-
   // if the query builder was able to build a query actually execute it.
   if (filter.query) {
     logger.debug(`Executing aggregation search over ${resourceType}s using query: ${JSON.stringify(filter.query)}`);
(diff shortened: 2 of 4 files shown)


sarahmcdougall (changes requested):
Yay great work on getting the pagination to work in the test server!

I don't have any specific changes to request, but I feel like it might be worth putting some (if not all) of the new pagination code into a helper function in `baseUtils` since the base search is 100+ lines of code now. Maybe we could make a helper function that takes in the `searchParams`, `resourceType`, host and base version from req/args, and then return a populated links array that we can set to `searchBundle.link`.

sarahmcdougall (approved):

PR #116, opened 2022-07-26
Fix evaluate-measure reportType option

mgramigna:
individual -> subject in all uses, docs, and tests.

Note that fqm-execution still requires a reportType of `individual` in its calculation options. This is fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,8 @@ Optional parameters include:
 
 - `practitioner`: practitioner for which the measure will be calculated
 
-Currently, `measure` and `lastReceivedOn` parameters are not supported by the test server. The `subject-list` `reportType` is not supported by the test server - only `subject` and `population` `reportTypes` are supported at this time.
+Currently, `measure` and `lastReceivedOn` parameters are not supported by the test server. The `subject-list` `reportType` is not supported by the test server - only `subject` and `population` `reportTypes` are supported at this time,
+which will generate `individual` and `summary` `MeasureReport`s respectively.
 
 To use, first POST a measure bundle into your database, then send a GET request to `http://localhost:3000/4_0_1/Measure/<your-measure-id>/$evaluate-measure` with the required parameters.
 
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ Optional parameters include:
 
 - `practitioner`: practitioner for which the measure will be calculated
 
-Currently, `measure` and `lastReceivedOn` parameters are not supported by the test server. The `subject-list` `reportType` is not supported by the test server - only `individual` and `population` `reportTypes` are supported at this time.
+Currently, `measure` and `lastReceivedOn` parameters are not supported by the test server. The `subject-list` `reportType` is not supported by the test server - only `subject` and `population` `reportTypes` are supported at this time.
 
 To use, first POST a measure bundle into your database, then send a GET request to `http://localhost:3000/4_0_1/Measure/<your-measure-id>/$evaluate-measure` with the required parameters.
 
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -245,9 +245,17 @@ const evaluateMeasure = async (args, { req }) => {
   // or using unsupported report type
   validateEvalMeasureParams(req.query);
 
-  return req.query.reportType === 'population'
-    ? evaluateMeasureForPopulation(args, { req })
-    : evaluateMeasureForIndividual(args, { req });
+  const { reportType, subject } = req.query;
+
+  // If reportType is not specified, default to 'subject', but
+  // only if the 'subject' parameter is also specificed
+  if (reportType === 'subject' || (reportType == null && subject != null)) {
+    logger.debug('Evaluating measure for individual');
+    return evaluateMeasureForIndividual(args, { req });
+  }
+
+  logger.debug('Evaluating measure for population');
+  return evaluateMeasureForPopulation(args, { req });
 };
 
 /**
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -137,7 +137,7 @@ Optional parameters include:
 
 - `practitioner`: practitioner for which the measure will be calculated
 
-Currently, `measure` and `lastReceivedOn` parameters are not supported by the test server. The `subject-list` `reportType` is not supported by the test server - only `individual` and `population` `reportTypes` are supported at this time.
+Currently, `measure` and `lastReceivedOn` parameters are not supported by the test server. The `subject-list` `reportType` is not supported by the test server - only `subject` and `population` `reportTypes` are supported at this time.
```
lmd59, line 140: Because subject is not an actual MeasureReport type, should we specify here that the `subject` parameter corresponds with an individual measure report?

src/services/measure.service.js
```diff
@@ -337,7 +337,7 @@ const evaluateMeasureForIndividual = async (args, { req }) => {
     measurementPeriodEnd: req.query.periodEnd
   });
 
-  const { periodStart, periodEnd, reportType = 'individual', subject, practitioner } = req.query;
```
lmd59, line 340: This made me investigate default report types a little bit, and it should probably be addressed higher in this file (in the main `evaluateMeasure` handling?), but we should be following these defaults:
`If not specified, a default value of subject will be used if the subject parameter is supplied, otherwise, population will be used`

src/util/operationValidationUtils.js
```diff
@@ -37,11 +37,11 @@ function validateEvalMeasureParams(query) {
     }
   }
 
-  if (query.reportType === 'individual') {
+  if (query.reportType === 'subject') {
     const subjectReference = query.subject.split('/');
     if (subjectReference.length > 1 && subjectReference[0] !== 'Patient') {
       throw new BadRequestError(
-        `For report type 'individual', subject reference may only be a Patient resource of format "Patient/{id}".`
+        `For report type 'subject', subject reference may only be a Patient resource of format "Patient/{id}".`
```
lmd59, line 44:
```suggestion
        `For report type parameter 'subject', subject reference may only be a Patient resource of format "Patient/{id}".`
```

lmd59: Just a suggestion to help avoid confusion between a report type parameter versus an actual report type.

test/util/operationValidationUtils.test.js
```diff
@@ -152,15 +152,15 @@ describe('validateEvalMeasureParams', () => {
...
+      query: { reportType: 'subject', periodStart: '2019-01-01', periodEnd: '2019-12-31' }
     };
     try {
       validateEvalMeasureParams(MISSING_SUBJECT_REQ.query);
       expect.fail('validateEvalMeasureParams failed to throw error for missing subject param');
     } catch (e) {
       expect(e.statusCode).toEqual(400);
       expect(e.issue[0].details.text).toEqual(
-        `Must specify subject for all $evaluate-measure requests with reportType: individual`
+        `Must specify subject for all $evaluate-measure requests with reportType: subject`
```
lmd59, line 163:
```suggestion
        `Must specify subject for all $evaluate-measure requests with reportType parameter: subject`
```

test/util/operationValidationUtils.test.js
```diff
@@ -185,30 +185,30 @@ describe('validateEvalMeasureParams', () => {
     }
   });
 
-  test('error thrown for individual $evaluate-measure with non-Patient reference subject', () => {
+  test('error thrown for subject $evaluate-measure with non-Patient reference subject', () => {
```
lmd59, line 188:
```suggestion
  test('error thrown for subject reportType parameter for $evaluate-measure with non-Patient reference subject', () => {
```

test/util/operationValidationUtils.test.js
```diff
@@ -185,30 +185,30 @@ describe('validateEvalMeasureParams', () => {
...
+        reportType: 'subject',
         periodStart: '2019-01-01',
         periodEnd: '2019-12-31',
         subject: 'Group/testGroup'
       }
     };
     try {
       validateEvalMeasureParams(INDIVIDUAL_REQ.query);
-      expect.fail('validateEvalMeasureParams failed to throw error for individual report with non-Patient subject');
+      expect.fail('validateEvalMeasureParams failed to throw error for subject report with non-Patient subject');
```
lmd59, line 199:
```suggestion
      expect.fail('validateEvalMeasureParams failed to throw error for subject reportType parameter with non-Patient subject');
```

test/util/operationValidationUtils.test.js
```diff
@@ -185,30 +185,30 @@ describe('validateEvalMeasureParams', () => {
...
     } catch (e) {
       expect(e.statusCode).toEqual(400);
       expect(e.issue[0].details.text).toEqual(
-        `For report type 'individual', subject reference may only be a Patient resource of format "Patient/{id}".`
+        `For report type 'subject', subject reference may only be a Patient resource of format "Patient/{id}".`
```
lmd59, line 203:
```suggestion
        `For report type parameter 'subject', subject reference may only be a Patient resource of format "Patient/{id}".`
```

src/services/measure.service.js
mgramigna: Good catch. Just pushed a fix. New logic is:

1. If BOTH reportType AND subject are missing, use population
2. If reportType is missing and subject is NOT missing, use subject

This adheres to the spec as I understand it now

lmd59 (approved):

PR #117, opened 2022-07-26
Fix updating resource logic

mgramigna:
# Summary

we were using findOneAndUpdate which didn't allow you to remove properties from resources. We wanted `replaceOne` instead. Rest is the same

## New behavior

Removing property during a PUT request from an existing resource now works

## Code changes

* Change to replaceOne
* Add test case for it
* Remove crazy test case in measure.service that had insane side effects (not sorry)

# Testing guidance

Sent PUT for a new resource. Should upsert as before

Send a PUT to modify that resource to add a property to it. Should work

Send a PUT to modify that resource to modify an existing property. Should work

Send a PUT to modify that resource to *delete* an existing property. Should work

### DIFF
--- a/test/services/measure.service.test.js
+++ b/test/services/measure.service.test.js
@@ -22,8 +22,6 @@ const { initialize } = require('../../src/server/server');
 const { SINGLE_AGENT_PROVENANCE } = require('../fixtures/provenanceFixtures');
 const testParamResource = require('../fixtures/fhir-resources/parameters/paramNoExportResource.json');
 
-const updateMeasure = { resourceType: 'Measure', id: 'testMeasure', name: 'anUpdate' };
-
 let server;
 
 describe('measure.service', () => {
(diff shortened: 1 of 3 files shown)


test/services/measure.service.test.js
```diff
@@ -69,19 +67,6 @@ describe('measure.service', () => {
...
-        .put('/4_0_1/Measure/testMeasure')
-        .send(updateMeasure)
-        .set('Accept', 'application/json+fhir')
-        .set('content-type', 'application/json+fhir')
-        .set('x-provenance', JSON.stringify(SINGLE_AGENT_PROVENANCE))
-        .expect(200)
-        .then(response => {
-          expect(response.headers.location).toBeDefined();
-        });
-    });
```
sarahmcdougall, line 83: this is a little sus don't ya think :eyes:

sarahmcdougall (approved):
:cowboy-eyes:

PR #118, opened 2022-07-27
Ecqm-content-r4-2021 upload script

natjoe4:
# Summary
Replacing connectathon upload script with ecqm-content-r4-2021 bundles since these are more commonly used nowadays

## New behavior
Renamed `connectathon-upload` script to `upload-bundles`. The script now, when run with no args, pulls all the measure bundles from the `ecqm-content-r4-2021` repo. An optional `dirPath` argument can be provided to specify upload from a different directory, and a `searchPattern` arg can also be included to specify a regex to filter out bundles for upload by file name.

## Code changes
 - `getConnectathonBundles.js` -> `uploadPremadeBundles.js`
 - `connnectathon-upload` -> `upload-bundles`
 - Default `upload-bundles` run now pulls from `ecqm-content-r4-2021` repo
 - Added option to include regex as an arg to filter out files by filename for upload

# Testing guidance
 - Follow instructions in the README.md to upload ecqm-content-r4-2021 bundles and ensure all bundles upload
 - Try running the upload-bundles script with and without `dirPath` and `searchPattern` args and ensure that the results are as expected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 node_modules
 .vscode/
 connectathon
+ecqm-content-r4-2021
 coverage
 docker_ssl_setup.sh
 tmp
--- a/README.md
+++ b/README.md
@@ -103,7 +103,8 @@ The following `npm` commands can be used to set up the database:
 - `npm run db:setup` creates collections for all the valid FHIR resource types
 - `npm run db:delete` deletes all existing collections in the database
 - `npm run db:reset` runs both of the above, deleting all current collections and then creating new, empty collections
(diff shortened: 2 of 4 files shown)


mgramigna (changes requested):
Ok cool looks/works. I have a few weird things that existed on main but we should just resolve here:

1. Why are we logging an error when res.resource.resourceType === 'Measure'? we do this on main too. This makes it so that errors don't show up during first upload but do after the second if you immediately re-run the script. We want to ignore duplicate key errors for measures too.. right?
2. Can we plz switch the bundlePromises logic to a `for (const filePath of bundleFiles)` loop? We can use await within that (the individual bundle entry uploads can stay as a `Promise.all`). This way we process one bundle at a time, which is much more preferable for large directories (JS can run out of heap space trying to do them all at once in a huge directory cause it needs to store everything in memory at once). If we do this, can you also put a ``console.log(`Uploading ${filePath}`)`` statement to see what the current bundle we're processing is? That'll be helpful I think

mgramigna (approved):
beautiful

PR #128, opened 2022-10-31
Ignore modelinfo dependency and update DB_HOST

lmd59:
# Summary
Update functionality to ignore modelinfo file dependencies because these are not necessary for calculation and are not formatted properly for use of the provided canonical url.
Update issue with mongo connection client introduced with node 17 and thus affecting node updates from 16->18/lts.

## New behavior

Basic requests should work against measures provided in the [ecqm-content-r4-2021 repo](https://github.com/cqframework/ecqm-content-r4-2021).

## Code changes

Environment variable update to be able to connect with database while using node 18.
Update to bundleUtils.js to ignore modelinfo dependency.

# Testing guidance

'npm run test`
It would be good to check main against node 18 before merging this code to make sure it is necessary. Otherwise, check with node 16 and node 18. Make sure that normal functionality continues to work. Sample requests as described [here](https://confluence.mitre.org/display/TAC/Sample+Requests) should work.

### DIFF
--- a/.env
+++ b/.env
@@ -1,6 +1,6 @@
 SERVER_HOST = localhost
 SERVER_PORT = 3000
-DB_HOST = localhost
+DB_HOST = 127.0.0.1
 DB_PORT = 27017
 DB_NAME = deqm-test-server
 KEYCLOAK_VERSION = 15.0.2
--- a/.env.test
+++ b/.env.test
@@ -1,4 +1,4 @@
-DB_HOST = localhost
+DB_HOST = 127.0.0.1
 DB_PORT = 27017
 DB_NAME = deqm-test-server-test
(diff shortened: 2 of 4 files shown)


sarahmcdougall (approved):
Changes look good to me. Tested out the fhir model info exclusion and that works - it's nice not to get errors for that anymore! This branch works for both node v16 and node v18. I tried testing the main branch on node v18 but the tests were hanging, so it seems like these changes are necessary to get things working with node v18. 

I think it should be fine to merge with the failing checks since the failures are coming from the test annotations, which we have seen before. The actual check for the unit tests passes and I'm not getting any issues locally

sarahmcdougall (approved):

PR #131, opened 2022-12-21
Unit test overhaul part 1- using jest-mongodb

elsaperelli:
# Summary
Prior to this PR, our unit testing approach for mongodb operations involved the use of `.env.test`. In our [Measure Repository Service Repository](https://github.com/projecttacoma/measure-repository-service), to unit test mongodb operations we use [jest-mongodb](https://github.com/shelfio/jest-mongodb), a jest preset to run memory MongoDB memory server. The goal of this PR is to also use this jest present in our unit tests in this repository for making a mongo connection.

## New behavior
- Behavior should remain the same except for the unit tests. The unit tests that were previously making a mongo connection using the `.env.test` file should now use jest-mongodb.

## Code changes
- `jest-mongodb-config.js` - configures the in memory mongo database (see the jest-mongodb readme linked above).
- `jest.config.js` - moved the jest configuration from `package.json` here and added the jest-mongodb preset.
- `package.json` / `package-lock.json` - upgraded jest version, added jest-mongodb dependency, removed `--runInBand` from the test command.
- `src/database/connection.js` - modified this file so that it contains a `Connection` class, similar to the one in our Measure Repository service. This makes it easier to use jest-mongodb in our unit tests. References to things in this file had to change in other files, but the functionality should remain the same.
- The rest of the files mainly contain changes to references to the `connection.js` file now that it contains a Connection class. 
- `test/populateTestData.js` - `testSetup` now connects to the in memory mongo database through `await Connection.connect(global.__MONGO_URI__);` Two other functions were added (`createEmptyCollection` and `resourceTestSetup`) to test the resource count of an empty collection in `resourcecount.service.test.js`. This is not the best fix but the functionality works for now.
- The test files now use the modified `testSetup` function so many of the `beforeAll()`, `beforeEach()`
, `afterEach()`, and `afterAll()` calls had to be added or modified. 
- A lot of spaces were added in test files for readability and consistency.

# Testing guidance
- `npm run check` to ensure that all tests pass and there are no lint errors/warnings.
- `npm upload-bundles` and `npm start` to ensure that the server functionality also remains the same.
- Definitely look at the `beforeAll()` etc. calls in the test files. I had to change a lot of this around and had the most trouble with this so make sure that it makes sense and works how it is supposed to.

NOTE: I realize that this is not super high priority right now and there is still a lot that can be improved about these unit tests. I consider this just to be part 1 of a unit test overhaul for this repository.

### DIFF
--- a/jest-mongodb-config.js
+++ b/jest-mongodb-config.js
@@ -1,0 +1,12 @@
+module.exports = {
+  mongodbMemoryServerOptions: {
+    binary: {
+      verison: '6.0.1',
+      skipMD5: true
+    },
+    autoStart: false,
+    instance: {}
+  },
+
+  useSharedDBForAllJestWorkers: false
+};
--- a/jest-mongodb-config.js
+++ b/jest-mongodb-config.js
@@ -1,7 +1,7 @@
 module.exports = {
   mongodbMemoryServerOptions: {
     binary: {
-      verison: '6.0.1',
+      version: '6.0.1',
       skipMD5: true
     },
     autoStart: false,
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@ const { buildConfig } = require('./config/profileConfig');
 const { initialize } = require('./server/server');
 const childProcess = require('child_process');
 const os = require('os');
+require('./config/envConfig');
 
 const app = express();
 app.use(express.json({ limit: '50mb', type: 'application/json+fhir' }));
@@ -41,10 +42,11 @@ for (let i = 0; i < process.env.NDJSON_WORKERS; i++) {
 }
 
 const port = process.env.SERVER_PORT || 3000;
+const url = `mongodb://${process.env.DB_HOST}:${process.env.DB_PORT}`;
 
 server.listen(port, async () => {
   logger.info(`Starting the FHIR Server at localhost:${port}`);
-  await mongoUtil.client.connect();
+  await mongoUtil.Connection.connect(url);
   logger.info('Connected to database');
   if (process.env.VALIDATE === 'true') {
     await axios.put(`http://${process.env.VALIDATOR_HOST}:${process.env.VALIDATOR_PORT}/igs/hl7.fhir.us.qicore`);
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,4 @@
+module.exports = {
+  preset: '@shelf/jest-mongodb',
+  setupFiles: ['./test/globalSetup.js']
+};
--- a/package-lock.json
+++ b/package-lock.json
@@ -3996,9 +3996,9 @@
       "integrity": "sha512-QADzlaHc8icV8I7vbaJXJwod9HWYp8uCqf1xa4OfNu1T7JVxQIrUgOWtHdNDtPiywmFbiS12VjotIXLrKM3orQ=="
     },
     "node_modules/cookiejar": {
-      "version": "2.1.3",
-      "resolved": "https://registry.npmjs.org/cookiejar/-/cookiejar-2.1.3.tgz",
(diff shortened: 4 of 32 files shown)


jest-mongodb-config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  mongodbMemoryServerOptions: {
+    binary: {
+      verison: '6.0.1',
```
lmd59, line 4:
```suggestion
      version: '6.0.1',
```

src/index.js
```diff
@@ -41,10 +42,11 @@ for (let i = 0; i < process.env.NDJSON_WORKERS; i++) {
 }
 
 const port = process.env.SERVER_PORT || 3000;
+const url = `mongodb://${process.env.DB_HOST}:${process.env.DB_PORT}`;
```
lmd59, line 45: Is there a way that we can make this url setting an intelligent default as part of the mongoUtil.Connection class? (i.e. if it's not otherwise set, set it to this). I'm worried that we're doing a bit of repeating setting this across the application.

lmd59 (commented):
I'm not sure I understand the different approaches for afterAll(cleanUpTest) changes - we're sometimes adding, sometimes removing, sometimes using clearTMP. Maybe we can discuss a bit so I better understand the context! :-)

jest-mongodb-config.js
elsaperelli: I may never come back from this

elsaperelli:
> I'm not sure I understand the different approaches for afterAll(cleanUpTest) changes - we're sometimes adding, sometimes removing, sometimes using clearTMP. Maybe we can discuss a bit so I better understand the context! :-)

These are drove me a bit crazy so good question! Correct me if I am wrong, but I don't believe I added any afterAll(cleanUpTest) when that wasn't there already.

clearTMP was added as a function only for the `bulk.service.test.js` file. As you can see in the changes in `test/populateTestData.js`, `cleanUpTest()` used to be the only function- but now it is split between `cleanUpTest()` and `clearTMP()`. Since removing `-- runInBand`, the tests sort of run at the same time which causes problems because `cleanUpTest()` always cleared the tmp folder. Splitting the function into two separate functions allowed me to fix this because the tmp folder was only being cleared when it needed to be. Does that make sense? I would be happy to talk through all of this when we have a chance so I can better understand it myself.

src/index.js
elsaperelli: Not sure, but I will look into this!

lmd59:
> > I'm not sure I understand the different approaches for afterAll(cleanUpTest) changes - we're sometimes adding, sometimes removing, sometimes using clearTMP. Maybe we can discuss a bit so I better understand the context! :-)
> 
> These are drove me a bit crazy so good question! Correct me if I am wrong, but I don't believe I added any afterAll(cleanUpTest) when that wasn't there already.
> 
> clearTMP was added as a function only for the `bulk.service.test.js` file. As you can see in the changes in `test/populateTestData.js`, `cleanUpTest()` used to be the only function- but now it is split between `cleanUpTest()` and `clearTMP()`. Since removing `-- runInBand`, the tests sort of run at the same time which causes problems because `cleanUpTest()` always cleared the tmp folder. Splitting the function into two separate functions allowed me to fix this because the tmp folder was only being cleared when it needed to be. Does that make sense? I would be happy to talk through all of this when we have a chance so I can better understand it myself.

So I thought there were instances of adding, but looking back, I see the addition is actually moving a describe-level afterAll to a test-specific afterAll (which I missed the corresponding removal for!).
That does help clarify! Thank you for taking the time to explain. Now going back to look at the cleanupTest and clearTMP, I realize I misread some things looking at the diff, and your explanation cleared up the misread *and* the reason for the change. :-)

PR #135, opened 2023-01-30
Change periodStart and periodEnd params to optional for dataRequirements

natjoe4:
# Summary
Converted periodStart and periodEnd parameters to optional for $data-requirements operation. Also bumped fqm-execution dependency to 1.0.5

## New behavior
An error will no longer be thrown when periodStart and periodEnd parameters are omitted from a $data-requirements operation instead, each situation will be handled in the following manor: 

 - If a measurement period start and end calculation option are provided, this period will be used
 - If a measurement period start calculation option is provided with no end, a period with duration 1 year beginning at the provided start will be used
 - If a measurement period end calculation option is provided with no start, a period with duration 1 year ending at the provided end will be used
 - If no measurement period start or end calculation option is provided, the measurement period will be determined by the effective period of the FHIR Measure, using the same logic as above to handle missing start and end dates
 - If no measurement period start or end calculation option is provided and no effectivePeriod is found on the measure, no measurement period will be used, and all dateFilters will be excluded from the resulting library

## Code changes
 - Deleted validateDataRequirementsParams function and removed all calls to it
 - Updated README
 - Bumped fqm-execution to 1.0.5
 - Updated unit tests

# Testing guidance
 - `npm install`
 - `npm run db:reset`
 - `npm run upload-bundles`
 - `npm start`
 - Head over to insomnia and run the attached test suite
 - Ensure the output for each example fits the logic outlined above. Keep in mind that the effectivePeriod of ColorectalCancerScreeningsFHIR is 2022-01-01 - 2022-12-31
[dr-tests.json.zip](https://github.com/projecttacoma/deqm-test-server/files/10537152/dr-tests.json.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,10 +190,12 @@ Check out the [$care-gaps operation spec](https://build.fhir.org/ig/HL7/davinci-
 
 This operation retrieves all the data requirements for a given measure as a FHIR library.
 
-Required parameters include:
+Optional parameters for this function include:
 
 - `periodStart`: start of the measurement period
 - `periodEnd`: end of the measurement period
+
+If `periodStart` and `periodEnd` params are omitted, the measurement period for the operation will default to the `effectivePeriod` of the referenced FHIR Measure. If no `effectivePeriod` property is present, `dateFilters` will be excluded from the returned FHIR Library entirely.
 
 To use, first POST a measure bundle into your database, then send a GET request to `http://localhost:3000/4_0_1/Measure/<your-measure-id>/$data-requirements`.
 
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ Optional parameters for this function include:
 - `periodStart`: start of the measurement period
 - `periodEnd`: end of the measurement period
 
-If `periodStart` and `periodEnd` params are omitted, the measurement period for the operation will default to the `effectivePeriod` of the referenced FHIR Measure. If no `effectivePeriod` property is present, `dateFilters` will be excluded from the returned FHIR Library entirely.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -190,11 +190,13 @@ Check out the [$care-gaps operation spec](https://build.fhir.org/ig/HL7/davinci-
 
 This operation retrieves all the data requirements for a given measure as a FHIR library.
 
-Required parameters include:
+Optional parameters for this function include:
 
 - `periodStart`: start of the measurement period
 - `periodEnd`: end of the measurement period
 
+If `periodStart` and `periodEnd` params are omitted, the measurement period for the operation will default to the `effectivePeriod` of the referenced FHIR Measure. If no `effectivePeriod` property is present, `dateFilters` will be excluded from the returned FHIR Library entirely.
```
elsaperelli, line 198:
```suggestion
If `periodStart` and `periodEnd` parameters are omitted, the measurement period for the operation will default to the `effectivePeriod` of the referenced FHIR Measure. If no `effectivePeriod` property is present, `dateFilters` will be excluded from the returned FHIR Library entirely.
```
I actually don't know if this is necessary/what conventions are for READMEs haha

elsaperelli, line 198: Do we want to add a description of the logic for when only a periodStart or periodEnd is added? Or is that unnecessary

elsaperelli (changes requested):
Two small README comments that aren't super necessary, otherwise looks good!

elsaperelli (approved):
Lgtm!

sarahmcdougall (approved):
I was confused thinking the changes weren't coming through only to realize I didn't run `npm install` to get the fqm-exec updates 🤦‍♀️ but after npm installing everything looks good, and good call with updating the README

PR #153, opened 2024-06-10
New import

elsaperelli:
# Summary
This PR changes the workflow of the bulk $import operation. Previously, this server used the Ping and Pull approach, but now it uses the proposed bulk data import outlined in [this IG](file:///Users/eperelli/Downloads/proposed-bulk-import-ig/site/index.html). 

## New behavior
For more information on the previous Ping and Pull approach, refer to this repository's Wiki page [here](https://github.com/projecttacoma/deqm-test-server/wiki/Bulk-Data-Import). The main difference between that approach and the new approach is that the new approach utilizes already exported inputs rather than this server kicking off the export request. This simplifies the process by including the export ndjson urls in the Import Manifest in the request body of the $import operation rather than just the export url. 

## Code changes
- `package.json` / `package-lock.json` - removed `bulk-data-utilities` dependency.
- `dbOperations.js` - two new functions for storing and getting ndjson file failed outcomes, inclusion of the import manifest in the `bulkImportClient` object so that it may be included on the Import Result.
- `importWorker.js` - `executeNewImportWorkflow` replaces `executePingAndPull`.
- `ndjsonWorker.js` - ndjson failed outcomes are now stored as well.
- `bulkstatus.service.js` - **WORK IN PROGRESS** - the structure of the Import Result object has yet to be finalized.
- `import.service.js` - use `retrieveInputUrls`.
- `measure.service.js` - `bulkSubmitData` now follows the new `$import` workflow. This is okay for now because `$bulk-submit-data` may even go away.
- `exportUtils.js` - `retrieveInputUrls` now returns all ndjson fileUrls included in the Import Manifest
- `test` - deleted param files that are no longer used in tests, updated unit tests to follow new workflow

# Testing guidance
- `npm run check`
- Utilize the [bulk-export-server](https://github.com/projecttacoma/bulk-export-server) to get the export ndjson url files to include in the Import Manifest. This can be done by populating the bulk-export-server, kicking off the $export request, and then polling the bulkstatus location until ndjson file urls are returned. Make sure the `bulk-export-server` is running on a different port than `deqm-test-server` (`PORT=3001 npm start`).
- Spin up the `deqm-test-server`: `npm start`
- Send a GET request to `http://localhost:3000/4_0_1/$import` with an Import Manifest json object in the request body (defined in IG) that includes ndjson file urls from the `bulk-export-server`. I am happy to send examples, but I figured leaving this open would be better for testing to make sure that I interpreted the IG correctly.
- Make sure the $import kickoff returns a 202 status.
- Send a GET request to the content-location (bulkstatus polling location). Make sure this either returns in-progress or OK status with a Import Result object.

# Import Results
Note that the structure of the Import Results (`bulkstatus.service.js`) is still a work in progress.


src/util/exportUtils.js
```diff
@@ -2,84 +2,37 @@ const { BadRequestError } = require('./errorUtils');
...
-      console.warn('_typeFilter parameter already supplied in exportUrl. Omitting entries from parameter array');
-    } else {
-      // add type filters from parameters to exportUrl
-      exportUrl += `${exportUrl.includes('_type=') ? '&' : '?'}_typeFilter=${typeFilterString}`;
-    }
-  }
-  return exportUrl;
+const retrieveInputUrls = parameters => {
+  logger.debug(`Retrieving all input URLs from parameters: ${JSON.stringify(parameters)}`);
+  const inputParamArray = parameters.filter(param => param.name === 'input');
```
hossenlopp, line 11: I noticed that if I don't include the `Content-Type: application/fhir+json` header in the request, the code ends up here with `parameters` being undefined as if something didn't allow the body to pass when the content was wrong instead of responding that the content type was unexpected.

src/util/exportUtils.js
```diff
@@ -2,84 +2,37 @@ const { BadRequestError } = require('./errorUtils');
...
-  if (!exportType) {
-    return 'dynamic';
-  } else {
-    return exportType.valueCode;
+const checkInputUrlArray = inputParamArray => {
+  if (inputParamArray.length === 0) {
+    throw new BadRequestError('No inputUrl parameters were found.');
   }
+  inputParamArray.forEach(inputUrl => {
+    if (!inputUrl.part.find(p => p.name === 'url').valueUrl) {
```
hossenlopp, line 32: This is a bit insufficient to check for bad input such as: 
```
{
  "resourceType": "Parameters",
  "parameter": [
    {
      "name": "input",
      "part": [
        {
          "name": "url",
          "valueUrl": "http://localhost:3001/b4d9c2c8-2630-4b80-a9e3-16fe26156dae/Patient.ndjson"
        }
      ]
    },
    {
      "name": "input",
      "part": [
      ]
    }
  ]
}
```

This causes a 500 to be returned since it tried to fetch `valueUrl` on undefined. Perhaps do another check on if `find()` found the `url` part or not and report that back as bad requestion.

src/services/import.service.js
```diff
@@ -9,21 +9,19 @@ const logger = require('../server/logger');
...
   logger.debug(`Request body: ${JSON.stringify(req.body)}`);
   logger.debug(`Request params: ${JSON.stringify(req.params)}`);
 
   // ID assigned to the requesting client
-  const clientEntry = await addPendingBulkImportRequest();
+  const clientEntry = await addPendingBulkImportRequest(req.body);
 
-  const exportURL = retrieveExportUrl(req.body.parameter);
-  const exportType = retrieveExportType(req.body.parameter);
+  const inputUrls = retrieveInputUrls(req.body.parameter);
```
hossenlopp, line 20: This adds the bulk import request to the database before it has been validated to be a good request. Perhaps this should happen later after its been validated to be a good request.

src/util/exportUtils.js
elsaperelli: I am pretty sure this is fixed- can you confirm?

src/util/exportUtils.js
elsaperelli: I see! Maybe we can discuss this- I am not sure what level this error should be thrown? I am not sure if the spec is clear whether this should be an error thrown during the kick-off request or if it should be a transient error that occurs during the $import that we then see in the bulkstatus response (the latter is what we currently do).

src/services/import.service.js
elsaperelli: I think the code I changed in this file may make more sense and make this OBE but lmk!

src/util/exportUtils.js
hossenlopp: It looks like we already have a utility for this in [baseUtils.js](https://github.com/projecttacoma/deqm-test-server/blob/dc89180bf4b355e97db38cfa84471dd16db82e93/src/util/baseUtils.js#L32). Just have to use it in the `import.service.js` where the request is initially being handled with a simple `checkContentTypeHeader(req.headers);`

src/util/exportUtils.js
hossenlopp: Hmmm... now it says that it creates the import request just fine (which isn't good) and when you check the bulkstatus url you get the `batch-response` bundle with an OperationOutcome that says "Internal System Error: '$import' request not processed."

I was expecting this to return with a 400 error on the initial `$import` request with an indication of malformed import manifest.

hossenlopp: After discussing on slack, this is fine for the error to come back in the bulkstatus endpoint. But for this situation it should return an error indicating the issue with parsing the import manifest.

src/services/import.service.js
hossenlopp: Yep. This makes more sense now that the error is recorded for return on the bulk status endpoint.

hossenlopp (changes requested):
The $import requests if you give a non-existent file (one that would return 404) in the input. I'll look more closely at what's causing this to happen.

hossenlopp:
Additional error handling should be added to the `ndjsonWorker.js` file to catch any mongo or Axios errors (such as a 404) and probably note their outcome and either fail the whole import or just that single file and create the OperationOutcome for that ndjson.

Suggest wrapping the most of the `ndjson.process` handler with a try catch to handle any unforeseen errors. Potentially add try catch with re-throwing for the `retrieveNDJSONFromLocation` function if you want to make the error message more descriptive that it had to do with request for the data failing.

src/server/ndjsonWorker.js
```diff
@@ -39,7 +39,18 @@ ndjsonWorker.process(async job => {
...
+  let ndjsonResources;
+  try {
+    ndjsonResources = await retrieveNDJSONFromLocation(fileUrl);
+  } catch (e) {
+    const outcome = [`ndjson retrieval of ${fileUrl} failed with message: ${e.message}`];
+
+    await pushNdjsonFailedOutcomes(clientId, fileUrl, outcome);
+    await pushBulkFailedOutcomes(clientId, outcome);
+    process.send({ clientId, resourceCount: 0, successCount: 0 });
+    return;
```
hossenlopp, line 52:
```suggestion
    logger.info(`ndjson-worker-${process.pid}: failed to fetch ${fileName}`);
    return;
```
I noticed this didn't log when it completed its attempt of processing the file. Suggest adding a log output here for that.

hossenlopp (changes requested):
Almost there with this! Just a logger statement would help with seeing whats going on on the console with our testing.

hossenlopp (approved):
Looks good and makes good use of the ping and pull stuff that was already there.

lmd59 (approved):
Looks great and thanks for all of the hard work on this one! My only comment is certainly a nitpick, so feel free to ignore if you disagree:
There are comment references and function names that call this the "new" import workflow. In the end this will just be the import workflow, so should we remove the "new" from comments and in function name?

lmd59 (approved):
Looks great!

hossenlopp (approved):

PR #164, opened 2025-05-12
Updated $evaluate output to parameters resource 

zacharyrobin:
# Summary
Updated the $evaluate [Implementation Guide specification](https://build.fhir.org/ig/HL7/davinci-deqm/OperationDefinition-evaluate.html). Previously, the server returned a single MeasureReport directly. Now, the response is wrapped in a FHIR Parameters resource containing one or more Bundle resources, each of which includes one or more MeasureReports.

## New behavior
The changes work both for `reportType=subject` (individual) and `reportType=population` responses

## Code changes
Updated `evaluateMeasureForPopulation` and `evaluateMeasureForIndividual` to wrap results in one or more Bundles.

Added new utility function `wrapReportsInBundlesParameters` to generate the Parameters resource from grouped `MeasureReports`.

# Testing guidance
`npm run check`
`npm run db:reset`
`npm run upload-bundles --test`

Using insomnia/ postman or curl, test the endpoints using:
GET `http://localhost:3000/4_0_1/Measure/{measureId}/$evaluate?periodStart=2025-01-01&periodEnd=2025-12-31&reportType=subject&subject=Patient/{patientId}` 
and
GET `http://localhost:3000/4_0_1/Measure/{measureId}/$evaluate?periodStart=2025-01-01&periodEnd=2025-12-31&reportType=population`

### DIFF
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -278,7 +278,7 @@ const evaluateMeasureForPopulation = async (args, { req }) => {
     });
 
     logger.info('Successfully generated $evaluate report');
-    return wrapReportsInBundlesParameters(results);
+    return wrapReportsInBundlesParameters([results]);
   }
 };
 
--- a/src/services/measure.service.js
+++ b/src/services/measure.service.js
@@ -278,10 +278,7 @@ const evaluateMeasureForPopulation = async (args, { req }) => {
     });
 
     logger.info('Successfully generated $evaluate report');
-
-    // grouping measureReports by patient
-    const groupedMeasureReports = results.map(report => [report]);
-    return wrapReportsInBundlesParameters(groupedMeasureReports);
+    return wrapReportsInBundlesParameters(results);
   }
 };
 
@@ -329,36 +326,32 @@ const evaluateMeasureForIndividual = async (args, { req }) => {
     measurementPeriodEnd: periodEnd,
     reportType: 'individual'
   });
-  const groupedMeasureReports = [[results[0]]];
-  return wrapReportsInBundlesParameters(groupedMeasureReports);
+
+  return wrapReportsInBundlesParameters(results);
 };
 
 /**
  * Wraps groups of measureReports to bundles and wraps each Bundle in a parameter
- * @param {Array<Array<Object>>} groupedMeasureReports An array of arrays of
- * MeasureReports (one array per Bundle).
+ * @param {Array<Object>} measureReports An array of measureReports.
  * @returns {Object} A FHIR Parameters resource containing one parameter per Bundle.
  */
-const wrapReportsInBundlesParameters = groupedMeasureReports => {
-  const parameters = groupedMeasureReports.map(reportGroup => {
-    const bundle = {
-      resourceType: 'Bundle',
-      // TODO: do we need an id here?
-      type: 'collection',
-      entry: reportGroup.map(report => ({
-        resource: report
-      }))
-    };
+const wrapReportsInBundlesParameters = measureReports => {
+  const bundle = {
+    resourceType: 'Bundle',
+    type: 'collection',
+    entry: measureReports.map(report => ({
+      resource: report
+    }))
+  };
 
-    return {
-      name: 'return',
-      resource: bundle
-    };
-  });
+  const parameter = {
+    name: 'return',
+    resource: bundle
+  };
 
   return {
     resourceType: 'Parameters',
-    parameter: parameters
+    parameter: [parameter]
   };
 };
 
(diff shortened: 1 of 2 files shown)


src/services/measure.service.js
```diff
@@ -326,7 +329,37 @@ const evaluateMeasureForIndividual = async (args, { req }) => {
...
+ * Wraps groups of measureReports to bundles and wraps each Bundle in a parameter
+ * @param {Array<Array<Object>>} groupedMeasureReports An array of arrays of
+ * MeasureReports (one array per Bundle).
+ * @returns {Object} A FHIR Parameters resource containing one parameter per Bundle.
+ */
+const wrapReportsInBundlesParameters = groupedMeasureReports => {
+  const parameters = groupedMeasureReports.map(reportGroup => {
+    const bundle = {
+      resourceType: 'Bundle',
+      // TODO: do we need an id here?
```
zacharyrobin, line 346: Question: I currently am not adding an ID to the bundle, do we need an id here?

lmd59 (commented):
A couple of thoughts inline about handling the array of arrays bit, but let me know if you saw other guidance on how to organize the measure reports into bundles.
Two additional requests:

1. Can you update the testing guidance in the PR description to reflect testing with the $evalute endpoint? (like you could suggest the http endpoint, something like http://localhost:3000/4_0_1/Measure/{measure id}/$evaluate?periodStart=2025-01-01&periodEnd=2025-12-31&reportType=subject&subject=Patient/{patient id} and /Measure/{measure id}$evaluate?periodStart=2025-01-01&periodEnd=2025-12-31&reportType=population or you can recommend certain testing with the frontend app).
2. Currently, the README says:
> Check out the [$evaluate operation spec](https://hl7.org/fhir/us/davinci-deqm/2024Sep/OperationDefinition-evaluate.html) for more information. Note: The server does not currently return a Bundle, but updates will be made to reflect this latest OperationDefinition.

I would update this to something like (make sure to maintain the link):

> This operation returns a Parameters object with 0..* Bundles, each of which must contain at least one MeasureReport. Check out the [$evaluate operation spec](https://hl7.org/fhir/us/davinci-deqm/2024Sep/OperationDefinition-evaluate.html) for more information.

src/services/measure.service.js
```diff
@@ -326,7 +329,37 @@ const evaluateMeasureForIndividual = async (args, { req }) => {
...
     reportType: 'individual'
   });
-  return results[0];
+  const groupedMeasureReports = [[results[0]]];
+  return wrapReportsInBundlesParameters(groupedMeasureReports);
+};
+
+/**
+ * Wraps groups of measureReports to bundles and wraps each Bundle in a parameter
+ * @param {Array<Array<Object>>} groupedMeasureReports An array of arrays of
```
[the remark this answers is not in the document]
zacharyrobin: I was thinking it needed to be an Array<Array<Objects>> because in the [operationDef](https://build.fhir.org/ig/HL7/davinci-deqm/OperationDefinition-evaluate.html) it says that it "Returns [Parameters](https://www.hl7.org/fhir/parameters.html) resource, which contains one or more Bundles. These Bundles SHALL contain 1..* DEQM Individual MeasureReport, [...]". so I was thinking that its a Parameter with an array of bundles that each can have an array of measure reports. Not sure if this is correct though

src/services/measure.service.js
zacharyrobin: hoss agrees and says we dont need the ID

src/services/measure.service.js
```diff
@@ -212,7 +212,7 @@ const evaluateMeasure = async (args, { req }) => {
  * Evaluate measure for "population" report type
  * @param {Object} args the args object passed in by the user, includes measure id
  * @param {Object} req http request object
- * @returns {Object} FHIR MeasureReport with population results
+ * @returns {Object} Parameters resource containing one or more Bundles of MeasureReports.
```
lmd59, line 215:
```suggestion
 * @returns {Object} Parameters resource containing one Bundle with MeasureReports.
```
Just to match the other return definition more closely

src/services/measure.service.js
```diff
@@ -326,7 +326,33 @@ const evaluateMeasureForIndividual = async (args, { req }) => {
     measurementPeriodEnd: periodEnd,
     reportType: 'individual'
   });
-  return results[0];
+
+  return wrapReportsInBundlesParameters(results);
+};
+
+/**
+ * Wraps groups of measureReports to bundles and wraps each Bundle in a parameter
```
lmd59, line 334:
```suggestion
 * Wraps groups of measureReports in a bundle and wraps the Bundle in a parameter
```

src/services/measure.service.js
```diff
@@ -278,15 +278,15 @@ const evaluateMeasureForPopulation = async (args, { req }) => {
     });
 
     logger.info('Successfully generated $evaluate report');
-    return results;
+    return wrapReportsInBundlesParameters(results);
```
lmd59, line 281:
```suggestion
    return wrapReportsInBundlesParameters([results]);
```
The summary report for a population has a single MeasureReport result rather than an array of MeasureReport.

lmd59 (approved):

PR #169, opened 2025-09-03
submit data update for v5 spec

lmd59:
# Summary
Update server to handle submit-data according to the 5.0.0 specification: https://build.fhir.org/ig/HL7/davinci-deqm/OperationDefinition-submit-data.html

## New behavior
$submit-data expects to receive a parameters object with 1..* transaction bundles and responds with 1..* transaction-response bundles (unless there's a failure). Individual POST/PUTs within a given transaction bundle may fail (returning The "September 2025 Connectathon" folder has a document discussing some of the implementation questions that were brought up as a part of this task. 

## Code changes

- Remove improper $submit-data format endpoint
- Update bundle service to return a single Promise (Promise.all)
- Update measure service to handle appropriate multi-bundle input and structure response into a Parameters
- Update validation to check for 1..* bundles in
- Update fixtures and tests, including updates to test to clear out the database in between db-changing submit-data tests

# Testing guidance

- `npm run check`
- Test with deqm-test-kit module
- May also play around with insomnia testing using input similar to the fixtures in the tests


zacharyrobin (approved):
No changes from me. looks great!

PR #172, opened 2025-09-21
TypeScriptifying Initial Pass

hossenlopp:
# Summary

Sets up the scaffolding for the initial TypeScriptification of the project. All files were renamed to `.ts` with `//@ts-nocheck` added. Also updates CI and DockerFiles to base everything off Node 20.x.

## New behavior

Now you can `npm run build` to build the project. Run the TS directly with `npm run start`. Run the JS with `node build/index.js`.

## Code changes

It is now "TypeScript". Some changes were made to paths in `require` calls to deal with the changeover to TS. Linting issues were also resolved. Coverage configuration was updated to properly include the `src` and `test` directory and ignore the `src/scripts`. This will result in a drop in coverage percentage because the previous configuration was only counting files reached by unit tests instead of all relevant files in the project.

# Testing guidance

Run unit tests. Spot check functionality when executing with `npm run start` and `node build/index.js`.


lmd59 (commented):
I'd like to do a bit more testing, but looking through the code, I found one small comment. And then I also think we should make some updates to the `package.json` file in this PR. Scripts are a bit broken, looking for specific js files or have a js search pattern.  And I think we need to update the `main` field.

PR #184, opened 2026-06-03
Make data requirements FHIR queries patient specific

elsaperelli:
# Summary
This PR adds a function that will be used later in the `$collect-data` workflow that takes a measure's data-requirements and uses patient information to make the queries patient-specific.

## New behavior
There should be no new behavior on the server itself, just the addition of the `collectUtils.ts` file and its functions.

## Code changes
- `src/util/collectDataUtils.ts` - new file for functions to make data-requirements patient-specific
- `test/util/collectDataUtils.test.ts` - associated unit test
- `tsconfig.json` - added "node" and "fhir" types to avoid editor errors

# Testing guidance
- `npm run check`

### DIFF
--- a/src/util/collectDataUtils.ts
+++ b/src/util/collectDataUtils.ts
@@ -1,0 +1,37 @@
+import { getMeasureBundleFromId } from './bundleUtils';
+import { Calculator, DRCalculationOutput } from 'fqm-execution';
+
+const PATIENT_ID_CONTEXT_TOKEN = '{{context.patientId}}';
+
+function populatePatientIdContext(dataRequirements: DRCalculationOutput, patientId: string): DRCalculationOutput {
+  return {
+    ...dataRequirements,
+    results: {
+      ...dataRequirements.results,
+      dataRequirement: dataRequirements.results.dataRequirement?.map(dataRequirement => ({
+        ...dataRequirement,
+        extension: dataRequirement.extension?.map(extension => ({
+          ...extension,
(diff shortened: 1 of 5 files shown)


src/util/collectDataUtils.ts
```diff
@@ -0,0 +1,40 @@
...
+            ? {
+                ...extension,
+                valueString: extension.valueString.split(PATIENT_ID_CONTEXT_TOKEN).join(patientId)
+              }
+            : extension
+        )
+      }))
+    }
+  };
+}
```
lmd59, line 15: Sorry I didn't see/suggest this earlier, but it's probably overkill to copy the whole object for a string replacement in the extension. Is there a reason we're not just manipulating the existing object instead of re-building the whole thing?
Something like this...
```suggestion
function populatePatientIdContext(dataRequirements: DRCalculationOutput, patientId: string): DRCalculationOutput {
  dataRequirements.results.dataRequirement?.forEach(dataRequirement => {
    dataRequirement.extension?.forEach(extension => {
      if (extension.valueString) {
        extension.valueString = extension.valueString.replaceAll(PATIENT_ID_CONTEXT_TOKEN, patientId);
      }
    });
  });

  return dataRequirements;
}
```

lmd59 (commented):
One more suggestion! Otherwise looks good.

lmd59 (approved):
Looks good! 🎉

PR #185, opened 2026-06-12
Collect data

lmd59:
# Summary
Adds initial invited pull $collectdata which uses the dataEndpoint field to pull data from a Clinical Data Repository.

## New behavior
Only supports one patient/measure for now, extremely basic validation, and synchronous data collection. Other support is expected to be added in future tickets. Has been tested with https://r4.smarthealthit.org used as the Clinical Data Repository.

## Code changes

- Adds endpoint
- Small cleanup in bundle service
- Added functionality in bundle, collect data, and validation utils to support url-based fetching and expanded parameters
- Main updates are in measure service to implement the new endpoint, parameter handling, data fetching, and Parameters/Bundle/MeasureReport final output creation
- Updates to README and tests

Added TODOs are functionality expected in future tasking or functionality noted in the bottom of this PR description.

# Testing guidance

- `npm run check`
- `npm run db:reset`
- `npm run upload-bundles`
- `npm run start`
- Use Insomnia to POST to http://localhost:3000/4_0_1/Measure/$collect-data

Example parameters:
```
{
"resourceType": "Parameters",
"parameter": [
   {
   	"name": "periodStart",
   	"valueDate": "2023-01-01"
   },
   {
   	"name": "periodEnd",
   	"valueDate": "2023-12-31"
   },
   {
   	"name": "measureUrl",
   	"valueCanonical": "http://ecqi.healthit.gov/ecqms/Measure/BreastCancerScreeningsFHIR"
   },
    {
      "name": "subject",
      "valueReference":{
				"reference": "Patient/d4fb3bba-73a9-4b82-a0bc-678d47f386b4"
			} 
    },
	{
      "name": "dataEndpoint",
      "resource": {
				"resourceType": "Endpoint",
				"id": "example",
				"status": "active",
				"connectionType": {
					"system": "http://terminology.hl7.org/CodeSystem/endpoint-connection-type",
					"code": "hl7-fhir-rest"
				},
				"payloadType": [
					{
						"coding": [
							{
								"system": "http://hl7.org/fhir/resource-types",
								"code": "Parameters"
							}
						]
					}
				],
				"payloadMimeType": [
					"application/fhir+json"
				],
				"address": "https://r4.smarthealthit.org"
			}
    }
  ]
}
```
Notes:

- Our data-requirements functionality (out of fqm-execution) does not produce a Patient resource data requirement by default. This seems like a potential oversight we should consider fixing.
- Tried updating default measure upload to use 2025 qicore bundles, but there appears to be some issues with dependent library urls, so I backed that update out. I believe this is a known issue with those bundles?
- It might be better to collate a transaction bundle for each patient and POST resources together with reference replacement (ala `const tb = createTransactionBundleClass('4_0_1'); tb.addEntryFromResource(testPatient, 'POST');`). But this could be the wrong approach if the clinical data queries result in any shared general data, like Medication resources. General data would be re-posted (duplicated) to the same server across different patient requests. I'm not sure whether or not dataRequirements generally pull that kind of data at this point (but it could theoretically be necessary for measure calculation). On the other hand, if you don't do a PUT-based approach (as we are currently doing in this PR), the receiving server could have id conflicts when receiving data from different dataEndpoints. We could do a more complex approach that identifies general/shared data, but that might be out of scope for the current task.


README.md
```diff
@@ -240,6 +241,22 @@ To use, first POST a measure bundle into your database, then send a GET request
 
 Check out the [$data-requirements operation spec](https://www.hl7.org/fhir/measure-operation-data-requirements.html) for more information.
 
+#### `$collect-data`
+
```
elsaperelli, line 245: Optional suggestion: may want to add that we only support one measure/one patient at the moment, but also this will change soonish so not totally necessary, you also kind of say this when you mention the contents of the returned Parameters resource.

test/fixtures/testDataRequirementsOutput.json
```diff
@@ -0,0 +1,58 @@
...
+            "valueString": "/Coverage?type=1,2,3&subscriber=Patient/{{context.patientId}}"
+          }
+        ]
+      },
+      {
+        "type": "Encounter",
+        "codeFilter": [
+          {
+            "path": "type",
+            "valueSet": "http://exmaple.com/ValueSet/exampleVS"
```
elsaperelli: LOL this is not your fault because this was in my code 😆 looks like I misspelled example if you don't find fixing that for me :)

src/services/measure.service.ts
```diff
@@ -142,6 +150,118 @@ const dataRequirements = async (args, { req }) => {
...
+    id: uuidv4(),
+    measure: measure.url?.includes('|') ? measure.url : `${measure.url}|${measure.version}`, //canonical measure/version
+    period: { start: periodStart, end: periodEnd },
+    status: 'complete',
+    type: 'data-collection',
+    subject: { reference: subject },
+    date: new Date().toISOString(),
+    reporter: { reference: 'Organization/deqm-test-server' },
+    meta: {
+      profile: ['http://hl7.org/fhir/us/davinci-deqm/StructureDefinition/datax-measurereport-deqm']
```
elsaperelli: Probably not a huge deal, but do we want to use the new IG one: "http://hl7.org/fhir/uv/deqm/StructureDefinition/deqm-dataexchangemeasurereport"?

elsaperelli (changes requested):
Just a few small comments but looks great!

elsaperelli (approved):
Lgtm!

PR #186, opened 2026-06-15
Replace urls as references not as values

ptran265:
# Summary
Differentiate when a url is being used as a reference versus when it's being used as a value instead of replacing all instances of the url

## New behavior
`replaceReferences` now replaces old `ResourceType/id` and `fullUrl` values only when they are used in a `reference` field, and does not replace those same strings when they are used as a value

## Code changes
-  Import fhir/r4 (I was getting errors throughout the whole file without this)
- New replaceReferences replacement algorithm to look for  `reference` properties for replacing instead of replacing all instances
- Added test

# Testing guidance
1. Check out https://github.com/cqframework/ecqm-content-cms-2025
2. Clear (npm run db:reset) and start (npm run start) the deqm-test-server 
3. Create an Insomnia POST request to url: http://localhost:3000/4_0_1/ 
4. Under the Body tab's dropdown, select file
5. Choose the mat CMS130 measure transaction bundle json file from the CMS 2025 repo: bundles/mat/CMS130FHIR-R2-MeasureExport/CMS130FHIR-v0.4.000-FHIR.json
6.Set header Content-Type to application/json+fhir
7. Send request (should receive back 200 OK and transaction-response bundle body)
8. Look at Measure collection to see what the `url` field is for the Measure added to the database. Confirm that it is https://madie.cms.gov/Measure/CMS130FHIRColorectalCancerScreening
9. Run tests (npm run test) and confirm all is passing 

(Previous bug is that the url is `[https://madie.cms.gov/Measure/\](https://madie.cms.gov/Measure/){some uuid}` rather than the original `https://madie.cms.gov/Measure/CMS130FHIRColorectalCancerScreening`)

### DIFF
--- a/src/util/bundleUtils.ts
+++ b/src/util/bundleUtils.ts
@@ -6,6 +6,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { findResourceById, findOneResourceWithQuery } from '../database/dbOperations';
 import logger from '../server/logger';
 import { Document, Filter } from 'mongodb';
+import * as fhir4 from 'fhir/r4';
 
 /*
  Some connectathon bundles currently contain incorrect url references from the main library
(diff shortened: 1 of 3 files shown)


src/util/bundleUtils.ts
```diff
@@ -6,6 +6,7 @@ import { v4 as uuidv4 } from 'uuid';
 import { findResourceById, findOneResourceWithQuery } from '../database/dbOperations';
 import logger from '../server/logger';
 import { Document, Filter } from 'mongodb';
+import * as fhir4 from 'fhir/r4';
```
lmd59: I think we've established that we don't need this line because of the tsconfig types, but we should work together to figure out how to get your IDE to recognize that.

src/util/bundleUtils.ts
```diff
@@ -241,18 +242,20 @@ export function replaceReferences(entries: any[]) {
 
   let entriesStr = JSON.stringify(entries);
   const postEntries = entries.filter(e => e.isPost);
+  const referenceProperty = (reference: string) => JSON.stringify({ reference }).slice(1, -1);
```
lmd59: Maybe a quick comment on this line or even pull this out into a separate function so that you can function comment it since it's a bit core to the below functionality?

src/util/bundleUtils.ts
```diff
@@ -241,18 +242,20 @@ export function replaceReferences(entries: any[]) {
...
   // For each POST entry, replace existing reference across all entries
   postEntries.forEach(e => {
     logger.debug(`Replacing referenceIds for entry: ${JSON.stringify(e)}`);
+    const newReference = `${e.resource.resourceType}/${e.newId}`;
     // Checking fullUrl and id in separate replace loops will prevent invalid ResourceType/ResourceID -> urn:uuid references
     if (e.oldId) {
-      const idRegexp = new RegExp(`${e.resource.resourceType}/${e.oldId}`, 'g');
-      entriesStr = entriesStr.replace(idRegexp, `${e.resource.resourceType}/${e.newId}`);
+      entriesStr = entriesStr
+        .split(referenceProperty(`${e.resource.resourceType}/${e.oldId}`))
```
lmd59: Any reason not to use `replaceAll` instead of split/join?

src/util/bundleUtils.ts
ptran265: Pulled it out to a separate function in [3eedf51](https://github.com/projecttacoma/deqm-test-server/pull/186/commits/3eedf51ceac480faf7d0dc92dd4c59e33c3f6ca4)

src/util/bundleUtils.ts
ptran265: That works too! I changed it in [3eedf51](https://github.com/projecttacoma/deqm-test-server/pull/186/commits/3eedf51ceac480faf7d0dc92dd4c59e33c3f6ca4)

lmd59 (approved):

PR #188, opened 2026-07-07
Update evaluate

elsaperelli:
# Summary
This PR updates the `$evaluate` operation to follow the [DEQM Universal Realm STU1 Definition for $evaluate](https://hl7.org/fhir/uv/deqm/2026May/en/OperationDefinition-evaluate.html).

## New behavior
- `measureId` is no longer supported
- `measureUrl` is required
- `practitioner` optional parameter is now `reporter` (we only support `reporter` as Practitioner reference right now, functionality should remain the name)
- Backwards compatibility allows for both `population`/`summary` and `subject`/`individual` report types

## Code changes
- `src/config/profileConfig.ts` - removes `Measure/[id]/$evaluate` support and replaces reference URL
- `src/queue/execQueue.ts` / `src/server/execWorker.ts` - uses `measureUrl` instead of `measureId`
- `src/services/measure.service.ts` - use `measureUrl` instead of `measureId`, replace `practitioner` with `reporter`
- `src/util/operationValidationUtils.ts` - updates to evaluate parameters validation function
- `test/fixutres/` - add url to fixtures
- `test/queue/execQueue.test.ts` - change `measureId` to `measureUrl`
- `test/server/execWorker.test.ts` - change `getMeasureBundleFromId` to `getMeasureBundleFromUrl`
- `test/services/measure.service.test.ts` / `test/util/operationValidationUtils.test.ts` - update to use `measureUrl` and `reporter`
- `README.md` - readme updates

# Testing guidance
- `npm run check`
- `npm run start`
- Test with the new test group in deqm-test-kit! 

# Additional Updates

## New behavior
- Support for all subject/reportType combinations as allowed by the spec
- Adjusts validation to recognized more unsupported parameters and allow any subject/reportType combination

## Code changes
- Updates measure service and patientUtils to gather initial patient ids from several sources (groups from either subjectGroup, subject Group reference, all patients) in combination with other filters (practitioner). Organizes result information appropriately to support reportType/subject combinations.
- Updates validation (adds some group validation to measureUtils for better shared
- Refactors supporting helpers for the measure service into measureUtils to clean up measure service and simplify testing
- Updates and adds tests (removes tests that disallow subject/reportType combinations)
- README updates

## Testing guidance
- `npm run check`
- `npm run db:reset`
- `npm run upload-bundles ../bundles/dqm-content-qicore-2025/bundles/measure` (depending on bundle location)
- `npm run start`

Run the Insomnia suite with the following caveats:
- After "create Group" request, copy the Group's id from the Location response header and use that for the Group reference in the body of the following "$evaluate Group Summary" request
- Before the "(limit db) $evaluate All Individual" request, make sure you have a limited database to create a response that doesn't overwhelm Insomnia or the heap:
- `npm run db:reset`
- `npm run upload-bundles ../bundles/dqm-content-qicore-2025/bundles/measure/CMS122FHIRDiabetesAssessGT9Pct` (depending on bundle location)
- `npm run start`
- Run the "(limit db) $evaluate All Individual" request

Insomnia tests:
[evaluateExpansion.yaml](https://github.com/user-attachments/files/30425076/evaluateExpansion.yaml)


src/services/measure.service.ts
```diff
@@ -379,32 +379,29 @@ const evaluateMeasure = async (args, { req }) => {
...
   // or using unsupported report type
-  validateEvalMeasureParams(query, args.id);
+  validateEvalMeasureParams(query);
 
   const { reportType, subject } = query;
 
   // If reportType is not specified, default to 'subject', but
   // only if the 'subject' parameter is also specified
-  if (reportType === 'subject' || (reportType == null && subject != null)) {
+  if (reportType === 'subject' || reportType === 'individual' || (reportType == null && subject != null)) {
```
lmd59: There's a bit of a subtlety here in the updated description of the reportType parameter. `a default value of summary will be used if the subject parameter is supplied and is a Group resource`, so we can't just check that subject is not null - we have to check the type of the resource of the subject reference and make sure it's not a group.

src/util/operationValidationUtils.ts
```diff
@@ -136,10 +133,12 @@ export function validateEvalMeasureParams(query: QueryObject, expectedId: string
     }
   }
 
-  if (query.practitioner && typeof query.practitioner === 'string') {
-    const practitionerReference = query.practitioner.split('/');
+  if (query.reporter && typeof query.reporter === 'string') {
+    const practitionerReference = query.reporter.split('/');
     if (practitionerReference.length !== 2 || practitionerReference[0] !== 'Practitioner') {
-      throw new BadRequestError(`practitioner may only be a Practitioner resource of format "Practitioner/{id}".`);
+      throw new BadRequestError(
```
lmd59, line 79: We should probably send a `NotImplementedError` instead for practitionerReference[0] === 'PractitionerRole' or practitionerReference[0] === 'Organization' since those would not be a bad request.

src/services/measure.service.ts
```diff
@@ -481,32 +477,30 @@ const evaluateMeasureForPopulation = async (args, query) => {
...
     const dataReq = await Calculator.calculateDataRequirements(measureBundle, {
       measurementPeriodStart: query.periodStart,
       measurementPeriodEnd: query.periodEnd
     });
 
-    const { periodStart, periodEnd, subject, practitioner } = query;
+    const { periodStart, periodEnd, subject, reporter } = query;
     let patientBundle;
-    if (practitioner) {
+    // does this change to reporter?
```
lmd59: Probably remove this line?

src/services/measure.service.ts
lmd59: I believe that the subject may also be a Group reference, so this is incomplete. This may require a bit more reworking throughout the function to allow for running an individual report on each of multiple patients (from a subject Group reference, subjectGroup, or whole server).

src/services/measure.service.ts
```diff
@@ -481,32 +477,30 @@ const evaluateMeasureForPopulation = async (args, query) => {
     const allResults = await Promise.all(resultsPromises);
 
     logger.info('Successfully generated $evaluate reports');
-    // an array of summary reports, one for each measure
```
lmd59, line 484: Is this comment no longer true? Trying to figure out why we're getting rid of some explanatory comments across the PR.

src/services/measure.service.ts
```diff
@@ -481,32 +477,30 @@ const evaluateMeasureForPopulation = async (args, query) => {
...
+
     return wrapReportsInBundlesParameters([allResults]);
```
lmd59, line 322: `allResults` is an array of summary reports, one for each measure. When we send it as a single nested array into `wrapReportsInBundlesParameters`, that function creates a single bundle. The updated STU1 description of the return is `The result of this operation is a Bundle for each input subject`. I think we either need to be putting all summary reports in a separate bundle for each patient (even though it's the same for each one), or we need clarification on the spec for summary reports.

src/config/profileConfig.ts
```diff
@@ -78,29 +78,17 @@ const buildConfig = () => {
...
-              route: '/:id/$evaluate',
-              method: 'POST',
-              reference: 'http://hl7.org/fhir/us/davinci-deqm/OperationDefinition/evaluate'
-            },
             {
               name: 'evaluateMeasure',
               route: '/$evaluate',
               method: 'GET',
-              reference: 'http://hl7.org/fhir/us/davinci-deqm/OperationDefinition/evaluate'
+              reference: 'http://hl7.org/fhir/uv/deqm/OperationDefinition/evaluatee'
```
lmd59:
```suggestion
              reference: 'http://hl7.org/fhir/uv/deqm/OperationDefinition/evaluate'
```

src/server/execWorker.ts
```diff
@@ -55,7 +55,7 @@ async function getMeasureBundle(measureId: string, periodStart: string, periodEn
   // load from mongo
   logger.info(`exec-worker-${process.pid}: Loading ${cacheLabel} from mongo.`);
   await mongoUtil.client.connect();
-  const measureBundle = await getMeasureBundleFromId(measureId);
+  const measureBundle = await getMeasureBundleFromUrl(measureUrl);
```
lmd59, line 58: Out of scope for this ticket, but just an FYI that I'm going to put a recommendation to do this a little differently (closer to the [bulk-export-server implementation](https://github.com/projecttacoma/bulk-export-server/blob/27dcf8096d786c44719e86bf27742bb64ae4a675/src/util/mongo.controller.js#L38)) in another ticket.

src/services/measure.service.ts
```diff
@@ -515,15 +509,15 @@ const evaluateMeasureForIndividual = async (args, query) => {
...
       const practitionerQuery = {
         id: patientId,
-        ...getResourceReference('generalPractitioner', practitioner)
+        ...getResourceReference('generalPractitioner', reporter)
       };
 
       const patient = await findOneResourceWithQuery(practitionerQuery, 'Patient');
       if (patient) {
-        patientBundle = await getPatientDataCollectionBundle(patient.id, dataReq.results.dataRequirement);
+        patientBundle = getPatientDataCollectionBundle(patient.id, dataReq.results.dataRequirement);
```
lmd59: Why are we removing the await?

src/util/operationValidationUtils.ts
```diff
@@ -41,26 +41,23 @@ function paramPresent(parameters: QueryObject, param: string) {
  * @param {Object} query query from http request object
  * @param {string} expectedId an id passed from the url arguments
  */
-export function validateEvalMeasureParams(query: QueryObject, expectedId: string) {
-  const REQUIRED_PARAMS = ['periodStart', 'periodEnd'];
+export function validateEvalMeasureParams(query: QueryObject) {
+  const REQUIRED_PARAMS = ['periodStart', 'periodEnd', 'measureUrl'];
   // currently only supports measureId as the identifier
-  const UNSUPPORTED_PARAMS = ['lastReceivedOn', 'measureIdentifier', 'measureUrl', 'measure', 'measureResource'];
+  const UNSUPPORTED_PARAMS = ['lastReceivedOn'];
```
lmd59: Not critical, but pretty easy to update this list to include all parameters from the evaluate operation spec that we do not support.

lmd59 (changes requested):
I haven't looked through quite everything yet, but I wanted to put up comments now. I think we need a good chunk of changes on this to align with spec, and I'm guessing we might need to make changes to the corresponding test kit.

src/config/profileConfig.ts
elsaperelli: LOL

src/util/operationValidationUtils.ts
elsaperelli: This is what I was looking for but was too lazy to find it lol...thank you!

src/util/operationValidationUtils.ts
lmd59: Addressed in https://github.com/projecttacoma/deqm-test-server/pull/188/commits/984869b46fe9975b941cff74da40975f410d5fc3

src/services/measure.service.ts
lmd59: Addressed in https://github.com/projecttacoma/deqm-test-server/pull/188/commits/984869b46fe9975b941cff74da40975f410d5fc3

src/services/measure.service.ts
lmd59: Addressed in https://github.com/projecttacoma/deqm-test-server/pull/188/commits/984869b46fe9975b941cff74da40975f410d5fc3

src/services/measure.service.ts
lmd59: Request for clarification is here: https://jira.hl7.org/browse/FHIR-57893

src/services/measure.service.ts
```diff
@@ -379,74 +244,47 @@ const evaluateMeasure = async (args, { req }) => {
...
-  } else {
-    if (query.practitioner) {
-      patientIds = await findResourceIdsWithQuery(
-        getResourceReference('generalPractitioner', query.practitioner),
-        'Patient'
+  if (query.reporter) {
+    const patients = await filterPatientByPractitionerFromIds(subjectIds, query.reporter);
+    if (patients.length === 0) {
+      throw new BadRequestError(
+        `The given subject has no patients that reference the given practitioner, ${query.reporter}`
```
elsaperelli: This is not a huge deal, but this error message shows up if the `subject` is a single Patient reference, which is potentially a little confusing if the user sees that the given Patient subject has no patients? Is there maybe a better error message to throw here?

src/services/measure.service.ts
```diff
@@ -482,94 +320,62 @@ const evaluateMeasureForPopulation = async (args, query) => {
...
+    query.measureUrl && Array.isArray(query.measureUrl)
+      ? await Promise.all(query.measureUrl.map(async m => await getMeasureBundleFromUrl(m)))
+      : [await getMeasureBundleFromUrl(query.measureUrl)];
+
+  let patientIds = [];
+  if (query.reporter) {
+    const patients = await filterPatientByPractitionerFromIds(subjectIds, query.reporter);
+    if (patients.length === 0) {
+      throw new BadRequestError(
+        `The given subject has no patients that reference the given practitioner, ${query.reporter}`
```
elsaperelli: Same comment as above.

elsaperelli (commented):
Just a few small comments, but looks great!!!

src/services/measure.service.ts
lmd59:
```suggestion
        `No provided subject patient(s) reference the given practitioner, ${query.reporter}`
```

src/services/measure.service.ts
lmd59:
```suggestion
        `No provided subject patient(s) reference the given practitioner, ${query.reporter}`
```

lmd59 (approved):
Clearing my previous review!

